PR #30, opened 2020-10-02
Add tests for k8sfiles.go

### LINKED ISSUE #28: Test the k8s metadata loaders.
# Description
This package looks in the source folder for k8s yaml files and loads the filepaths into the plan.
It parses the yaml files it finds to check if they are k8s resources.
It also transforms the files into runtime objects during translate phase.

# How to get started
For this issue you should start with some simple tests that creates temporary directories and calls the functions on it.
Next step would be to make slight changes to the files in the directory and create a separate subtest for each scenario.
Some scenarios that are good to test:
- Function called on an empty directory.
- Directory we don't have permission to read.
- Directory with only non yaml files.
- Directory with some non k8s yaml files.
- Directory with some k8s yamls and other non k8s files.

# How to add unit tests
Some guidelines:
- Look at the other tests in the package/project and write similar tests.
- Use subtests to test different paths through the function.
- Don't worry about testing every single path through the function. Focus on common use cases and failure modes.

Some helpful resources on how to write unit tests in Go:
- https://blog.golang.org/subtests
- https://youtu.be/hVFEV-ieeew

# Code to be tested
https://github.com/konveyor/move2kube/blob/731b16a14cd33dd864e51eac3f7945f35266b18f/internal/metadata/k8sfiles.go#L36-L80

vovapi:
Add tests for `metadata.K8sFilesLoader`

I added test directories in `testdata/k8s/` folder.

I covered all the cases described in #28:

> - Function called on an empty directory.
> - Directory we don't have permission to read.
> - Directory with only non yaml files.
> - Directory with some non k8s yaml files.
> - Directory with some k8s yamls and other non k8s files.

I brought new dependency: https://github.com/stretchr/testify. It is great tool set for testing. I used `github.com/stretchr/testify/suite.Suite` for organising subtests and setting testing environment before each subtest (which reduced copy-pasting). I also uses  `github.com/stretchr/testify/assert.Assertions` (embedded in suite.Suite) to write more clean error and value assertions.

I made all tests to pass, however I think think that loader should return error on bad permissions, while it is not (see TODO).
Looking for your suggestions on what use cases should return error.

Closes #28


ashokponkumar:
Thanks@vovapi. @HarikrishnanBalagopal Can you please review the commit and work with @vovapi. Also, look at why travis is failing

HarikrishnanBalagopal:
@vovapi Thanks for the PR!
Yes the library seems to be very useful for writing tests.

`Looking for your suggestions on what use cases should return error.`
Since the project is still in `alpha` we are trying to be as user friendly as possible.
For that reason it tries to continue processing to best effort even when there are errors.

`I made all tests to pass, however I think think that loader should return error on bad permissions, while it is not (see TODO).`
In this case I think the function could be written to return the error instead of ignoring it.

ashokponkumar:
Thanks @vovapi for your valuable commit. Please do continue to contribute to the project. We will take out an action item on your TODO.

HarikrishnanBalagopal:
Meant to merge and close.

PR #36, opened 2020-10-05
K8sFilesLoader: return error on bad directory permissions

### LINKED ISSUE #31: Don't ignore error and more idiomatic code.
# Description
1. As mentioned in the TODO https://github.com/konveyor/move2kube/blob/master/internal/metadata/k8sfiles_test.go#L77
we should not continue when the directory has incorrect permissions and we can't read it.

2. The functions we are testing https://github.com/konveyor/move2kube/blob/master/internal/metadata/k8sfiles.go#L36-L80 may not be the most idiomatic way to write them. For example: we have a `continue` and an `else` block.

# Changes to be made
Return with the error in the if block.
https://github.com/konveyor/move2kube/blob/584a35dcd4a23a809edaf0e9cd70080c08bad364/internal/metadata/k8sfiles.go#L39-L42

Change the test to expect the error.

Change the functions to be more idiomatic go.

vovapi:
K8sFilesLoader now handles bad directory permissions error and returns it.

GetFilesByExt, GetFilesByName are fixed and return bad file permissions error when it occurs on root search directory.
This makes possible to uncomment respective tests and remove TODOs: 
https://github.com/konveyor/move2kube/blob/c9b5d0f8497ccf1dc61dd90879da9ad8b427967a/internal/common/utils_test.go#L75-L85
https://github.com/konveyor/move2kube/blob/c9b5d0f8497ccf1dc61dd90879da9ad8b427967a/internal/common/utils_test.go#L130-L140

In addition to that changes K8sFilesLoader to be more idiomatic: removes unnecessary `else` block after `continue` in `for` loop

Removes TODO for bad permission test introduced in https://github.com/konveyor/move2kube/pull/30
Closes #31

### DIFF
--- a/internal/metadata/k8sfiles.go
+++ b/internal/metadata/k8sfiles.go
@@ -17,7 +17,9 @@ limitations under the License.
 package metadata
 
 import (
+	"errors"
 	"io/ioutil"
+	"syscall"
 
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
@@ -38,6 +40,9 @@ func (i K8sFilesLoader) UpdatePlan(inputPath string, plan *plantypes.Plan) error
 
 	files, err := common.GetFilesByExt(inputPath, []string{".yml", ".yaml"})
 	if err != nil {
+		if errors.Is(err, syscall.EACCES) { // if err is permission denied error then return it
+			return err
+		}
 		log.Warnf("Unable to fetch yaml files and recognize k8 yamls : %s", err)
 	}
 	for _, path := range files {
--- a/internal/metadata/k8sfiles.go
+++ b/internal/metadata/k8sfiles.go
@@ -56,9 +56,9 @@ func (i K8sFilesLoader) UpdatePlan(inputPath string, plan *plantypes.Plan) error
 		if err != nil {
 			log.Debugf("ignoring file %s since serialization failed", path)
 			continue
-		} else {
-			plan.Spec.Inputs.K8sFiles = append(plan.Spec.Inputs.K8sFiles, relpath)
 		}
+
+		plan.Spec.Inputs.K8sFiles = append(plan.Spec.Inputs.K8sFiles, relpath)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


ashokponkumar:
@HarikrishnanBalagopal PTAL

vovapi:
One more thing: in order participate in Hacktoberfest your project needs to put "hacktoberfest" topic to your repo:
Check out https://hacktoberfest.digitalocean.com/hacktoberfest-update

vovapi:
One more thing: in order participate in Hacktoberfest your project needs to put "hacktoberfest" topic to your repo:
Check out https://hacktoberfest.digitalocean.com/hacktoberfest-update

ashokponkumar:
> One more thing: in order participate in Hacktoberfest your project needs to put "hacktoberfest" topic to your repo:
> Check out https://hacktoberfest.digitalocean.com/hacktoberfest-update

Thanks @vovapi for the tip and for your continued valuable contribution. Have updated the topics with the tag. Do let us know your feedback, and continue to contribute.

internal/metadata/k8sfiles.go
```diff
@@ -38,6 +40,9 @@ func (i K8sFilesLoader) UpdatePlan(inputPath string, plan *plantypes.Plan) error
 
 	files, err := common.GetFilesByExt(inputPath, []string{".yml", ".yaml"})
 	if err != nil {
+		if errors.Is(err, syscall.EACCES) { // if err is permission denied error then return it
```
HarikrishnanBalagopal, line 43: I think it's best to always return here. @ashokponkumar what do you think? previously we were just warning and continuing https://github.com/konveyor/move2kube/blob/master/internal/metadata/k8sfiles.go#L41

HarikrishnanBalagopal: Or at least add `IsNotExist` to the list of errors which cause an early return
https://golang.org/pkg/os/#IsNotExist

Other than that it LGTM

ashokponkumar: Yes, since GetFilesByExt returns an error only if there is an error in the base input path, the fileslist will anyway be empty. So we can return early in all cases.

HarikrishnanBalagopal: @vovapi can you change it just return early if there is an error? No need to check if it `EACCESS`

ashokponkumar: @vovapi Can you possibly remove this check and update the PR?

ashokponkumar:
@HarikrishnanBalagopal Go ahead and merge the PR. Please use 'Rebase and merge'.

Issue #35, opened 2020-10-03
Refactor string normalization function and test it.


HarikrishnanBalagopal:
# Description
This package normalizes certain strings in the intermediate representation.
It strips any whitespace and quotations/commas from each container's environment variables.

Two things need to be done:
1. The package is not written in the most idiomatic way.
2. The package needs unit tests.

Some examples of code that needs to be refactored:
- This should be `for _, scObj := range ir.Services {`
https://github.com/konveyor/move2kube/blob/master/internal/optimizer/normalizecharactersoptimizer.go#L34-L35
- This should use `TrimSpace` https://golang.org/pkg/strings/#TrimSpace
https://github.com/konveyor/move2kube/blob/master/internal/optimizer/normalizecharactersoptimizer.go#L40-L41
- This function shouldn't return an error since it isn't supposed to fail:
https://github.com/konveyor/move2kube/blob/master/internal/optimizer/normalizecharactersoptimizer.go#L60-L64

# How to get started
For this issue you should start with some simple tests that creates simple IR objects and calls the function on them.
Next step would be to make slight changes to the IR objects and create a separate subtest for each scenario.
Some scenarios that are good to test:
- Function called on an IR with no services.
- An IR containing services that have no containers.
- An IR containing services and containers but the containers have no environment variables.
- An IR containing services and containers and all the environment variables are valid.
- An IR containing services and containers and some of the environment variables are invalid (containing spaces and quotes).
- An IR containing services and containers and some of the environment variables are invalid but their names contain the string `affinity`.

# How to add unit tests
Some guidelines:
- Look at the other tests in the package/project and write similar tests.
- Use subtests to test different paths through the function.
- Don't worry about testing every single path through the function. Focus on common use cases and failure modes.

Some helpful resources on how to write unit tests in Go:
- https://blog.golang.org/subtests
- https://youtu.be/hVFEV-ieeew

# Code to be tested
https://github.com/konveyor/move2kube/blob/c9b5d0f8497ccf1dc61dd90879da9ad8b427967a/internal/optimizer/normalizecharactersoptimizer.go#L32-L64

suramrit:
@HarikrishnanBalagopal Would like to get working on this!
Thanks for the great head start and background on the issue! 🚀

ShajithaMohammed:
@HarikrishnanBalagopal @ashokponkumar @suramrit Actually I have addressed this issue and ready to raise a PR,Forgot to sign-up. Apologies.
[...]

PR #42, opened 2020-10-09
Refactor string normalization function and add unit tests

ShajithaMohammed:


ashokponkumar:
@suramrit If you have already started work on #35, can you please check if you can merge your changes too to this or if you want to submit it as a PR on top of this change if it adds more coverage 0r has more fixes.

internal/optimizer/normalizecharactersoptimizer.go
```diff
@@ -31,34 +29,28 @@ type normalizeCharacterOptimizer struct {
...
-					if err == nil {
-						env.Name = tmpString
-					}
-					tmpString, err = stripQuotation(env.Value)
-					if err == nil {
-						env.Value = tmpString
-					}
+					env.Name = strings.TrimSpace(env.Name)
+					env.Value = strings.TrimSpace(env.Value)
+					tmpString := stripQuotation(env.Name)
```
HarikrishnanBalagopal, line 39: Don't use `tmpString` . Just assign directly `env.Name = stripQuotation(string.TrimSpace(env.Name))`
Same for `env.Value`

internal/optimizer/normalizecharactersoptimizer.go
```diff
@@ -31,34 +29,28 @@ type normalizeCharacterOptimizer struct {
...
+			ir.Services[i].Containers[j].Env = tmpEnvArray
 		}
-		ir.Services[k] = scObj
 	}
 	return ir, nil
 }
 
-func stripQuotation(inputString string) (string, error) {
+func stripQuotation(inputString string) string {
 	regex := regexp.MustCompile(`^[',"](.*)[',"]$`)
```
HarikrishnanBalagopal, line 53: This regex might need to be changed as well.
For now just add a comment with `TODO: check if regex is correct`

HarikrishnanBalagopal (changes requested):
Some minor comments. Overall a great PR! 👍

internal/optimizer/normalizecharactersoptimizer_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+	t.Run("strip matching single quotation marks from input string", func(t *testing.T) {
+		// Setup
+		inputString := "'testString'"
+		want := "testString"
+
+		// Test
+		actual := stripQuotation(inputString)
+		if !reflect.DeepEqual(actual, want) {
+			t.Fatal("Failed to get the expected properly. Expected:", want, "actual:", actual)
```
ShajithaMohammed, line 39: @HarikrishnanBalagopal Can we use  https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff instead of just printing expected and actual here as well? 
In order to maintain consistency throughout codebase and sometimes dealing with strings to find difference is also hectic.

HarikrishnanBalagopal: yes

suramrit:
> @suramrit If you have already started work on #35, can you please check if you can merge your changes too to this or if you want to submit it as a PR on top of this change if it adds more coverage 0r has more fixes.

I had started but only a little. Happy to let @ShajithaMohammed take charge and work on the issues, would not want to add confusion to their ongoing PR

HarikrishnanBalagopal (approved):
Thanks for making the changes. Please feel free to contribute to the other issues as well.

Issue #22, opened 2020-10-01
Improve Code coverage for unit tests


ashokponkumar:
Improve Code coverage for unit tests in the various packages of move2kube.

faabiosr:
I take care to improve the tests.

ShajithaMohammed:
I can help with unit tests for `optimize` package.

ashokponkumar:
sure @ShajithaMohammed . Please reach out to us if you need more specific questions. 

cc @HarikrishnanBalagopal Please let @ShajithaMohammed know if you have any specific tips.

HarikrishnanBalagopal:
@ShajithaMohammed Thanks for the PR. I will leave this issue open because it is about getting 100% (or as close to 100%) line coverage.

PR #46, opened 2020-10-10
Unit tests for optimize package

ShajithaMohammed:
As part of #22

### DIFF
--- a/internal/optimizer/imagepullpolicyoptimizer_test.go
+++ b/internal/optimizer/imagepullpolicyoptimizer_test.go
@@ -31,10 +31,9 @@ func TestImagePullPolicyOptimizer(t *testing.T) {
 
 	t.Run("IR with no services", func(t *testing.T) {
 		// Setup
-		p := plantypes.NewPlan()
-		ir := types.NewIR(p)
+		ir := getIRWithoutServices()
 		imagePullPolicyOptimizer := imagePullPolicyOptimizer{}
-		want := ir
+		want := getIRWithoutServices()
 
 		// Test
 		actual, err := imagePullPolicyOptimizer.optimize(ir)
@@ -48,17 +47,9 @@ func TestImagePullPolicyOptimizer(t *testing.T) {
 
 	t.Run("IR containing services that have no containers", func(t *testing.T) {
 		// Setup
-		svcname1 := "svcname1"
-		svcname2 := "svcname2"
-		svc1 := types.Service{Name: svcname1, Replicas: 2}
-		svc2 := types.Service{Name: svcname2, Replicas: 2}
-
-		p := plantypes.NewPlan()
-		ir := types.NewIR(p)
-		ir.Services[svcname1] = svc1
-		ir.Services[svcname2] = svc2
+		ir := getIRWithServiceAndWithoutContainers()
 		imagePullPolicyOptimizer := imagePullPolicyOptimizer{}
-		want := ir
+		want := getIRWithServiceAndWithoutContainers()
 
 		// Test
 		actual, err := imagePullPolicyOptimizer.optimize(ir)
@@ -88,7 +79,7 @@ func TestImagePullPolicyOptimizer(t *testing.T) {
 		p := plantypes.NewPlan()
 		ir := types.NewIR(p)
 		ir.Services[svcname1] = svc1
-		ir.Services[svcname1] = svc2
+		ir.Services[svcname2] = svc2
 
 		imagePullPolicyOptimizer := imagePullPolicyOptimizer{}
 		want := getIRWithImagePullPolicySetAsAlways()
@@ -107,7 +98,7 @@ func TestImagePullPolicyOptimizer(t *testing.T) {
 		// Setup
 		ir := getIRWithImagePullPolicySetAsAlways()
 		imagePullPolicyOptimizer := imagePullPolicyOptimizer{}
-		want := ir
+		want := getIRWithImagePullPolicySetAsAlways()
 
 		// Test
 		actual, err := imagePullPolicyOptimizer.optimize(ir)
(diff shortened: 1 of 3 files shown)


ashokponkumar:
@HarikrishnanBalagopal PTAL

internal/optimizer/imagepullpolicyoptimizer_test.go
```diff
@@ -0,0 +1,144 @@
...
+		svcname2 := "svcname2"
+		svc1 := types.Service{Name: svcname1, Replicas: 2}
+		svc2 := types.Service{Name: svcname2, Replicas: 4}
+		svc1.Containers = append(svc1.Containers, c1)
+		svc2.Containers = append(svc2.Containers, c2)
+
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		ir.Services[svcname1] = svc1
+		ir.Services[svcname1] = svc2
```
HarikrishnanBalagopal, line 91: Shouldn't this be `svcname2`?

internal/optimizer/imagepullpolicyoptimizer_test.go
```diff
@@ -0,0 +1,144 @@
...
+		if !cmp.Equal(actual, want) {
+			t.Fatalf("Failed to get the intermediate representation properly. Differences:\n%s", cmp.Diff(want, actual))
+		}
+	})
+
+	t.Run("IR containing services and containers with image pull policy already set as Always", func(t *testing.T) {
+		// Setup
+		ir := getIRWithImagePullPolicySetAsAlways()
+		imagePullPolicyOptimizer := imagePullPolicyOptimizer{}
+		want := ir
```
HarikrishnanBalagopal, line 110: See previous comment about assigning like this.

internal/optimizer/portmergeoptimizer_test.go
```diff
@@ -0,0 +1,378 @@
...
+
+func TestPortMergeOptimizer(t *testing.T) {
+	log.SetLevel(log.DebugLevel)
+
+	t.Run("IR with no services", func(t *testing.T) {
+		// Setup
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		portMergeOptimizer := portMergeOptimizer{}
+		want := ir
```
HarikrishnanBalagopal, line 36: See previous comment about assigning like this.

internal/optimizer/portmergeoptimizer_test.go
```diff
@@ -0,0 +1,378 @@
...
+		svc1.Containers = append(svc1.Containers, c1)
+		svc2.Containers = append(svc2.Containers, c2)
+
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		ir.Services[svcname1] = svc1
+		ir.Services[svcname2] = svc2
+
+		portMergeOptimizer := portMergeOptimizer{}
+		want := ir
```
HarikrishnanBalagopal, line 128: See previous comment about assigning like this.

internal/optimizer/replicaoptimizer_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+func TestReplicaOptimizer(t *testing.T) {
+	log.SetLevel(log.DebugLevel)
+
+	t.Run("IR with no services", func(t *testing.T) {
+		// Setup
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		replicaOptimizer := replicaOptimizer{}
+		want := ir
```
HarikrishnanBalagopal, line 35: See previous comment about assigning like this.

internal/optimizer/replicaoptimizer_test.go
```diff
@@ -0,0 +1,168 @@
...
+		svcname2 := "svcname2"
+		svc1 := types.Service{Name: svcname1, Replicas: 2}
+		svc2 := types.Service{Name: svcname2, Replicas: 2}
+
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		ir.Services[svcname1] = svc1
+		ir.Services[svcname2] = svc2
+		replicaOptimizer := replicaOptimizer{}
+		want := ir
```
HarikrishnanBalagopal, line 59: See previous comment about assigning like this.

internal/optimizer/replicaoptimizer_test.go
```diff
@@ -0,0 +1,168 @@
...
+		svcname2 := "svcname2"
+		svc1 := types.Service{Name: svcname1, Replicas: 4}
+		svc2 := types.Service{Name: svcname2, Replicas: 3}
+
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		ir.Services[svcname1] = svc1
+		ir.Services[svcname2] = svc2
+		replicaOptimizer := replicaOptimizer{}
+		want := ir
```
HarikrishnanBalagopal, line 107: See previous comment about assigning like this.

HarikrishnanBalagopal (changes requested):
All the comments are about 2 problems:
1. Assigning the input to the expected output with `:=` doesn't do a deep copy.
2. Some typos where `svcname1` should be `svcname2`.

Other than that LGTM.

internal/optimizer/imagepullpolicyoptimizer_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func TestImagePullPolicyOptimizer(t *testing.T) {
+	log.SetLevel(log.DebugLevel)
+
+	t.Run("IR with no services", func(t *testing.T) {
+		// Setup
+		p := plantypes.NewPlan()
+		ir := types.NewIR(p)
+		imagePullPolicyOptimizer := imagePullPolicyOptimizer{}
+		want := ir
```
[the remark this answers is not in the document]
ShajithaMohammed: In this case, can  we ignore `want` variable and replace tests with `if !cmp.Equal(actual, ir){}` ?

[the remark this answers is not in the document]
HarikrishnanBalagopal: @ShajithaMohammed in this case it might not make a difference but in general that would run into the same issue.
Can you please keep the expected output and actual output separate to avoid having to think about whether they are affecting each other?

HarikrishnanBalagopal (approved):
LGTM

PR #53, opened 2020-10-11
move CI routine to a make target

Jan200101:
make test, build, test-style are all called when running `make ci`


ashokponkumar:
Thanks for the PR @Jan200101 . Please feel free to make more enhancements and submit more PRs.

PR #71, opened 2020-10-16
Make script pass shellcheck. Avoid common bash pitfalls.

HarikrishnanBalagopal:

### DIFF
--- a/scripts/installdeps.sh
+++ b/scripts/installdeps.sh
@@ -16,7 +16,21 @@
 
 # initOS discovers the operating system for this system.
 
+[[ $DEBUG ]] || DEBUG='false'
+
 MOVE2KUBE_DEP_INSTALL_PATH="$PWD/bin"
+
+HAS_DOCKER="$(type docker &>/dev/null && echo true || echo false)"
+HAS_PACK="$(type pack &>/dev/null && echo true || echo false)"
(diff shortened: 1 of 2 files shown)


ashokponkumar:
@HarikrishnanBalagopal Can we add a line to get acknowledgement from user that these binaries will be installed? Also, warn them that if it is already installed, it might be updated?

scripts/installdeps.sh
```diff
@@ -14,28 +14,40 @@
...
+    OS="$(uname | tr '[:upper:]' '[:lower:]')"
+
+    case "$OS" in
+    # Minimalist GNU for Windows
+    mingw*) OS='windows' ;;
+    esac
+}
+
+initOS
+
```
ashokponkumar, line 28: Can we add a line to get acknowledgement from user that these binaries will be installed? Also, warn them that if it is already installed, it might be updated?

ashokponkumar (changes requested):

scripts/installdeps.sh
HarikrishnanBalagopal: `Can we add a line to get acknowledgement from user that these binaries will be installed?` Done

Not sure about warning them about updates. We are not updating anything.

ashokponkumar (approved):

PR #78, opened 2020-10-17
Fixed bug where we where merging/appending even when ther was an error. Added tests. Renamed method to be more consistent.

HarikrishnanBalagopal:


ashokponkumar:
@HarikrishnanBalagopal any idea why this patch reduced the code coverage quite significantly?

Issue #25, opened 2020-10-02
Add inbuilt support for popular managed cloud kubernetes services


ashokponkumar:
The task involves the following:

1. Setting up a terminal with the cluster in context for kubectl
1. Run `move2kube collect`
1. Copy the m2k_collect/clusters/<file>
1. Change name to <clustername>.yaml
1. In yaml, change metadata.name to clustername used above
1. Check in to https://github.com/konveyor/move2kube/tree/master/internal/metadata/clusters
1. Do `make generate`
1. Submit a pull request after committing with `git commit -s -m comment`

ashokponkumar:
@akhil-ghatiki If you have access to any kubernetes cluster from any of the cloud services, this might be a easy issue to fix, in case you are interested.

akhil-ghatiki:
I will check this out. I think I better work on this first than Add collector for Amazon ECS and Azure Container Service #43. (Because #43 requires lot of other exploration)

akhil-ghatiki:
@ashokponkumar I have set up the terminal with the cluster in context for kubectl for an aws eks. After running the `move2kube collect` the cluster folder is just empty. No yml got generated. 
<img width="1035" alt="Screenshot 2020-10-14 at 10 16 36 PM" src="https://user-images.githubusercontent.com/3032101/96020112-2e89bb80-0e6b-11eb-8a6d-0fecd936bb8e.png">

Is there any other configuration that has to be changed ? OR I afraid if #43 has to be fixed before this can be fixed.

ashokponkumar:
> @ashokponkumar I have set up the terminal with the cluster in context for kubectl for an aws eks. After running the `move2kube collect` the cluster folder is just empty. No yml got generated.
> <img alt="Screenshot 2020-10-14 at 10 16 36 PM" width="1035" src="https://user-images.githubusercontent.com/3032101/96020112-2e89bb80-0e6b-11eb-8a6d-0fecd936bb8e.png">
> 
> Is there any other configuration that has to be changed ? OR I afraid if #43 has to be fixed before this can be fixed.

@akhil-ghatiki #43 is about container service, and is not related to this issue. Can you please share the log output of `move2kube collect -v`?

akhil-ghatiki:
@ashokponkumar any more managed services support is required ? or is this issue good to close ?

ashokponkumar:
If you have access to managed openshift in any of them, we can use them too. If not, lets close it for now.

akhil-ghatiki:
[...]

PR #81, opened 2020-10-18
Add inbuilt support for Azure kubernetes service

akhil-ghatiki:
This addresses issue : https://github.com/konveyor/move2kube/issues/25


internal/metadata/clusters/azure_ks.yaml
```diff
@@ -0,0 +1,160 @@
+apiVersion: move2kube.io/v1alpha1
+kind: ClusterMetadata
+metadata:
+  name: AZURE-KS
```
ashokponkumar, line 4: @akhil-ghatiki Thanks. This is a great fix. Can we rename it as Azure-AKS, since Azure is not an acronym (hence non caps) and AKS will help users relate to it.

ashokponkumar (changes requested):

internal/metadata/clusters/azure_ks.yaml
akhil-ghatiki: Changed the name and pushed it.

internal/metadata/clusters/azure-aks.yaml
```diff
@@ -0,0 +1,160 @@
+apiVersion: move2kube.io/v1alpha1
```
ashokponkumar, line 1: One more small change @akhil-ghatiki We have change the domain name in the master branch. So for new providers, we can possibly use the one from master. But for this change if we can change the domain to "move2kube.konveyor.io" that should fix it.

akhil-ghatiki: Changed this and updated to new domain name

ashokponkumar (approved):

PR #90, opened 2020-10-21
complete test for checking cluster names

pabloloyola:


go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	github.com/whilp/git-urls v1.0.0
 	github.com/xrash/smetrics v0.0.0-20200730060457-89a2a8a1fb0b
 	golang.org/x/crypto v0.0.0-20201002170205-7f63de1d35b0
+	gopkg.in/yaml.v2 v2.3.0
```
ashokponkumar, line 38: Lets use gopkg.in/yaml.v3

internal/metadata/clustermdloader_test.go
```diff
@@ -17,6 +17,13 @@ limitations under the License.
 package metadata_test
 
 import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+
+	"gopkg.in/yaml.v2"
```
ashokponkumar, line 25: Lets use gopkg.in/yaml.v3

internal/metadata/clustermdloader_test.go
```diff
@@ -49,6 +56,60 @@ func TestUpdatePlan(t *testing.T) {
...
+			Metadata struct {
+				Name string `yaml:"name"`
+			} `yaml:"metadata"`
+		}
+
+		p := plantypes.NewPlan()
+		loader := metadata.ClusterMDLoader{}
+		cmMap := loader.GetClusters(p)
+		//fmt.Println(cmMap)
+		files, err := ioutil.ReadDir("clusters/")
```
ashokponkumar, line 72: We can reuse the below function to find all yaml files

https://github.com/konveyor/move2kube/blob/8afca8785e03fc85a6859b2ce6652916043cd016/internal/common/utils.go#L46

internal/metadata/clustermdloader_test.go
```diff
@@ -49,6 +56,60 @@ func TestUpdatePlan(t *testing.T) {
...
+			if re_yaml.MatchString(file.Name()) || re_yml.MatchString(file.Name()) {
+				yfiles = append(yfiles, file.Name())
+			}
+		}
+
+		// parse yaml files and collect the required info (medata-> names)
+		var metadata_names []string
+		fmt.Println("parsing yaml files:")
+		for _, yfile := range yfiles {
+			filename, _ := filepath.Abs("clusters/" + yfile)
```
ashokponkumar, line 92: Lets not use the same hard coded string in two places. Lets use a string constant and use in both places.

internal/metadata/clustermdloader_test.go
```diff
@@ -49,6 +56,60 @@ func TestUpdatePlan(t *testing.T) {
...
+			//fmt.Println(filename)
+			content, _ := ioutil.ReadFile(filename)
```
ashokponkumar, line 94: For reading an yaml you can use the below function

https://github.com/konveyor/move2kube/blob/8afca8785e03fc85a6859b2ce6652916043cd016/internal/common/utils.go#L179

internal/metadata/clustermdloader_test.go
```diff
@@ -49,6 +56,60 @@ func TestUpdatePlan(t *testing.T) {
...
+			//fmt.Println(test_struct.Metadata.Name)
+			metadata_names = append(metadata_names, test_struct.Metadata.Name)
+		}
+		fmt.Println("finish parsing yaml files")
+		fmt.Println(metadata_names)
+
+		//TODO: Read all .yaml files in internal/metadata/clusters, and find the value in metadata.name using say a regex
+
+		for _, clustername := range metadata_names {
+			if _, ok := cmMap[clustername]; !ok {
```
ashokponkumar, line 106: We should also possibly check the other way around, if there is a item in cmMap which is not in the metadata_names. It can even be two different tests.

ashokponkumar (changes requested):

PR #104, opened 2020-10-22
Fixed bug where order of items in the list was changing. Now we ignore the order while comparing.

HarikrishnanBalagopal:


ashokponkumar (changes requested):
Do we want to compare the other way around too?

PR #124, opened 2020-10-27
parameterizer basic tests -2

pabloloyola:

### DIFF
--- a/internal/parameterizer/parameterizer_test.go
+++ b/internal/parameterizer/parameterizer_test.go
@@ -1,0 +1,247 @@
+/*
+Copyright IBM Corporation 2020
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package parameterize_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	"github.com/google/go-cmp/cmp"
+
+	common "github.com/konveyor/move2kube/internal/common"
+	parameterize "github.com/konveyor/move2kube/internal/parameterizer"
+	"github.com/konveyor/move2kube/internal/types"
+	irtypes "github.com/konveyor/move2kube/internal/types"
+	plantypes "github.com/konveyor/move2kube/types/plan"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestParameterizer(t *testing.T) {
+	log.SetLevel(log.DebugLevel)
+
+	t.Run("1.IR with no services, no storage", func(t *testing.T) {
+
+		ir := getIRWithoutServices()
+		irAfter, err := parameterize.Parameterize(ir)
+
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		irAfter.TargetClusterSpec.Host = ""
+
+		if !cmp.Equal(ir, irAfter, cmpopts.EquateEmpty()) {
+			t.Fatalf("TODO explanation. Difference:\n%s", cmp.Diff(ir, irAfter))
+		}
+
+	})
+
+	/*
+		t.Run("2.IR containing services that have no containers", func(t *testing.T) {
+
+			ir := getIRWithServicesAndWithoutContainers()
+			irAfter, err := parameterize.Parameterize(ir)
+			if err != nil {
+				t.Fatal("Failed to parameterize. Error:", err)
+			}
+
+			// oracle
+			// if the service has no containers, then no changes will be done
+			// ir = irAfter
+			if !cmp.Equal(ir, irAfter, cmpopts.EquateEmpty()) {
+				t.Fatalf("TODO explanation. Difference:\n%s", cmp.Diff(ir, irAfter))
+			}
+
+		})
+	*/
+
+	t.Run("3.IR containing services with containers", func(t *testing.T) {
+
+		ir := getIRWithServicesAndContainers()
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+		fmt.Printf("%T", irAfter)
+		// oracle
+
+	})
+
+	t.Run("4.IR with no services , but storage, storage type is not PVCKind ", func(t *testing.T) {
+
+		// as the type is not PVCKind, the scMap will be empty
+		// -> no parameterizations -> before = after
+		ir := getIRWithStorageNotPVCKind()
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		// oracle
+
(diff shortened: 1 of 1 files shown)


internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+		irAfter, err := parameterize.Parameterize(ir)
+
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		irAfter.TargetClusterSpec.Host = ""
+
+		if !cmp.Equal(ir, irAfter, cmpopts.EquateEmpty()) {
+			t.Fatalf("TODO explanation. Difference:\n%s", cmp.Diff(ir, irAfter))
```
HarikrishnanBalagopal, line 51: Fix the TODOs. Add an explanation. It can be something as simple as `Failed to parameterize the IR properly...`

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+		ir := getIRWithoutServices()
+		irAfter, err := parameterize.Parameterize(ir)
+
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		irAfter.TargetClusterSpec.Host = ""
+
+		if !cmp.Equal(ir, irAfter, cmpopts.EquateEmpty()) {
```
HarikrishnanBalagopal, line 50: There is an issue with doing this. `ir` and `irAfter` are not independent.
Yes golang copies values when you pass them to function BUT, it does a shallow copy. Slices, maps, etc. aren't copied.
See https://play.golang.org/p/XvTTfaiFBy9

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+	*/
+
+	t.Run("3.IR containing services with containers", func(t *testing.T) {
+
+		ir := getIRWithServicesAndContainers()
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+		fmt.Printf("%T", irAfter)
```
HarikrishnanBalagopal, line 82: Don't using `fmt.Print` use the logging package `log`
You have already imported it.
This should probably be `log.Infof` or `log.Debugf`
Also why are we printing the type? `%T` We already know what type it is `github.com/konveyor/move2kube/internal/types.IR`

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+		t.Run("2.IR containing services that have no containers", func(t *testing.T) {
+
+			ir := getIRWithServicesAndWithoutContainers()
+			irAfter, err := parameterize.Parameterize(ir)
+			if err != nil {
+				t.Fatal("Failed to parameterize. Error:", err)
+			}
+
+			// oracle
+			// if the service has no containers, then no changes will be done
```
HarikrishnanBalagopal, line 66: Based on today's discussion please either keep or remove this test.
Don't leave it commented out, it is messing up the numbering of the other tests.

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		// oracle
+
+		irAfter.TargetClusterSpec.Host = ""
+		if !cmp.Equal(ir, irAfter, cmpopts.EquateEmpty()) {
+			t.Fatalf("TODO explanation. Difference:\n%s", cmp.Diff(ir, irAfter))
```
HarikrishnanBalagopal, line 101: Fix todo

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+		// scMap should be 1, we can parameterize
+		ir := getIRWithoutServices()
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		hostParam := "{{ .Release.Name }}-"
+
+		if irAfter.TargetClusterSpec.Host != hostParam {
```
HarikrishnanBalagopal, line 136: See previous comment about doing a full comparison.

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+	return ir
+}
+
+func getIRWithContainer() types.IR {
+
+	// Setup Containers
+	cname1 := "name1"
+	cnew1 := true
+	cont1 := types.NewContainer(plantypes.DockerFileContainerBuildTypeValue,
+		cname1,
```
HarikrishnanBalagopal, line 237: I think there's a formatting problem here since it isn't lined up.
Is `go fmt` being run when you save? Usually vscode will do that, if not please configure it to do so.

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func getIRWithServicesAndContainers() types.IR {
+	c1 := corev1.Container{
+		Name: "container-1",
+	}
+	c2 := corev1.Container{
+		Name: "container-2",
+	}
+	//c1.ImagePullPolicy = v1.PullAlways // not needed for the moment
```
HarikrishnanBalagopal, line 169: ?

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+	t.Run("3.IR containing services with containers", func(t *testing.T) {
+
+		ir := getIRWithServicesAndContainers()
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+		fmt.Printf("%T", irAfter)
+		// oracle
+
```
HarikrishnanBalagopal, line 84: Missing `cmp.Equal(actual, want)`.
We need to check the output being produced `actual` against some expected output `want`

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,247 @@
...
+		irAfter, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize. Error:", err)
+		}
+
+		// oracle:
+		paramSC := "{{ .Values.storageclass }}"
+		for _, j := range irAfter.Storages {
+			x := *j.PersistentVolumeClaimSpec.StorageClassName
+			if x != paramSC {
```
HarikrishnanBalagopal, line 119: Since the function takes the entire `ir` struct as input and produces a new `ir` struct, we should be doing full comparison with `cmp.Equal(actual, want)`. We need to make sure all the fields are as expected.

HarikrishnanBalagopal (changes requested):
Overall the basic structure is there.
Some changes are required like doing a full comparison with a separate `want` object.

ashokponkumar:
@HarikrishnanBalagopal please let me know when this pr is ready to be merged

internal/parameterizer/parameterizer_test.go
```diff
@@ -0,0 +1,250 @@
...
+		// scMap should be 1, we can parameterize
+		ir := getIRWithoutServices()
+		actual, err := parameterize.Parameterize(ir)
+		if err != nil {
+			t.Fatal("Failed to parameterize the IR properly. Error:", err)
+		}
+
+		hostParam := "{{ .Release.Name }}-{{ .Values.ingresshost }}"
+
+		if actual.TargetClusterSpec.Host != hostParam {
```
HarikrishnanBalagopal, line 149: check full IR

HarikrishnanBalagopal (approved):
The 6th test case is still not checking the full IR.
Other than that LGTM

PR #164, opened 2020-11-05
initial test for the default engine

pabloloyola:

### DIFF
--- a/internal/qaengine/defaultengine_test.go
+++ b/internal/qaengine/defaultengine_test.go
@@ -19,30 +19,35 @@ package qaengine
 import (
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
+	qatypes "github.com/konveyor/move2kube/types/qaengine"
+
+	"github.com/konveyor/move2kube/internal/common"
 	log "github.com/sirupsen/logrus"
 )
 
-func TestEngine(t *testing.T) {
+func TestDefaultEngine(t *testing.T) {
 	log.SetLevel(log.DebugLevel)
 
-	t.Run("1. test AddEngine", func(t *testing.T) {
-
-		//make sure the engines slice is empty
-		engines = nil
+	t.Run("Fetch the default answer", func(t *testing.T) {
 
 		e := NewDefaultEngine()
-		want := NewDefaultEngine()
 		AddEngine(e)
 
-		if len(engines) != 1 {
-			t.Fatalf("Engine was not added correctly to the engines slice. Length of engines slice: %d", len(engines))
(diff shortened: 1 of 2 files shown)


internal/qaengine/defaultengine_test.go
```diff
@@ -0,0 +1,53 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+func TestDefaultEngine(t *testing.T) {
+	log.SetLevel(log.DebugLevel)
+
+	t.Run("Fetch the default answer", func(t *testing.T) {
+
+		e := NewDefaultEngine()
+		AddEngine(e)
```
HarikrishnanBalagopal, line 34: Might have to set `engines = []Engine{}` here as well.
When you add more tests every test will need to have that.

internal/qaengine/defaultengine_test.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			log.Fatalf("Unable to create problem : %s", err)
+		}
+
+		problem, err = FetchAnswer(problem)
+		if err != nil {
+			log.Fatalf("Unable to fetch answer : %s", err)
+		}
+
+		if problem.Solution.Answer[0] != common.DefaultRegistryURL {
```
HarikrishnanBalagopal, line 48: Use the `GetXAnswer` functions:
https://github.com/konveyor/move2kube/blob/93ef9f7213a2c492c440bae2b38ba5d6e4ed46bf/types/qaengine/problem.go#L104-L145

Example:
https://github.com/konveyor/move2kube/blob/master/internal/customizer/registrycustomizer.go#L110-L113

internal/qaengine/engine_test.go
```diff
@@ -28,6 +28,9 @@ func TestEngine(t *testing.T) {
 
 	t.Run("1. test AddEngine", func(t *testing.T) {
 
+		//make sure the engines slice is empty
+		engines = nil
```
HarikrishnanBalagopal, line 32: This could be fine, but let's change it to `engine = []Engine{}` just in case.

internal/qaengine/defaultengine_test.go
```diff
@@ -0,0 +1,53 @@
...
+			t.Fatalf("Fetched answer was different from the default one. Fetched answer: %s", problem.Solution.Answer)
```
HarikrishnanBalagopal, line 49: Print the expected answer as well.

HarikrishnanBalagopal (changes requested):
Some minor changes.

HarikrishnanBalagopal (approved):
LGTM

PR #165, opened 2020-11-06
tests for other types of problems (on the default engine scenario)

pabloloyola:

### DIFF
--- a/internal/qaengine/defaultengine_test.go
+++ b/internal/qaengine/defaultengine_test.go
@@ -19,16 +19,16 @@ package qaengine
 import (
 	"testing"
 
-	qatypes "github.com/konveyor/move2kube/types/qaengine"
-
+	"github.com/google/go-cmp/cmp"
 	"github.com/konveyor/move2kube/internal/common"
+	qatypes "github.com/konveyor/move2kube/types/qaengine"
 	log "github.com/sirupsen/logrus"
 )
 
 func TestDefaultEngine(t *testing.T) {
 	log.SetLevel(log.DebugLevel)
 
-	t.Run("Fetch the default answer", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/qaengine/defaultengine_test.go
```diff
@@ -57,4 +57,135 @@ func TestDefaultEngine(t *testing.T) {
...
+			log.Fatalf("Unable to fetch answer : %s", err)
+		}
+
+		answer, err := problem.GetStringAnswer()
+		if err != nil {
+			log.Fatalf("Unable to get answer : %s", err)
+		}
+
+		if answer != def {
+			t.Fatalf("Fetched answer was different from the default one. Fetched answer: %v, expected answer: %v ",
```
HarikrishnanBalagopal, line 185: We can use %s because the answer and default are string

internal/qaengine/defaultengine_test.go
```diff
@@ -57,4 +57,135 @@ func TestDefaultEngine(t *testing.T) {
...
+
+	t.Run("Test NewMultilineInputProblem", func(t *testing.T) {
+
+		engines = []Engine{}
+		e := NewDefaultEngine()
+		AddEngine(e)
+
+		desc := "Test description"
+		context := []string{"Test context"}
+		def := "Option B"
```
HarikrishnanBalagopal, line 167: Since this is a multiline input problem, use a default with multiple lines. Example:
```
def := `line1
line2
line3`
```

HarikrishnanBalagopal (approved):
Some minor comments. Need to add a test for https://github.com/konveyor/move2kube/blob/master/types/qaengine/problem.go#L285-L288

Other than that LGTM.

HarikrishnanBalagopal (approved):
LGTM

PR #166, opened 2020-11-06
Update homepage- added asciinema embedding and updated links

Akash-Nayak:

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -14,7 +14,8 @@ tutorials_url: /tutorial
 gems:
 - jekyll-seo-tag
 github:
-  repository_url: https://konveyor.github.io/move2kube/
+  repository_url: https://www.github.com/konveyor/move2kube/
   repository_name: Move2Kube
   owner_url: https://konveyor.github.io/
   owner_name: Konveyor 
+  is_project_page: false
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -14,7 +14,8 @@ tutorials_url: /tutorial
 gems:
 - jekyll-seo-tag
 github:
-  repository_url: https://konveyor.github.io/move2kube/
+  repository_url: https://www.github.com/konveyor/move2kube/
   repository_name: Move2Kube
   owner_url: https://konveyor.github.io/
   owner_name: Konveyor 
+  is_project_page: false
```
ashokponkumar, line 21: Isn't it a project page?

ashokponkumar (changes requested):

_config.yml
Akash-Nayak: Actually, there were two GitHub buttons on the homepage. The `View on GitHub` button which comes up with the Cayman theme was taking to an incorrect link. I found this https://github.com/pages-themes/cayman/issues/29#issuecomment-339016299. So, to disable that button I have added this line.

ashokponkumar: Okay cool

ashokponkumar (approved):

PR #184, opened 2020-11-10
Change theme to just-the-docs

Akash-Nayak:
* Changing the theme of Move2Kube website from the [jekyll-theme-cayman](https://github.com/pietromenna/jekyll-cayman-theme) theme to the [just-the-docs](https://github.com/pmarsceill/just-the-docs) theme.
* The new theme gives additional functioning over the previous theme, like, enables Search, adds navigation bar to the left, etc.
* The website can be previewed at https://akash-nayak.github.io/move2kube/


ashokponkumar:
Are we using github ci engine? if not, we should remove .github folder.

Akash-Nayak:
> Are we using github ci engine? if not, we should remove .github folder.

No, I think we are not using it. So, I will remove the .github folder.

_config.yml
```diff
@@ -1,21 +1,121 @@
...
+# For technical reasons, this file is *NOT* reloaded automatically when you use
+# 'jekyll serve'. If you change this file, please restart the server process.
+
+# Site settings
+# These are used to personalize your new site. If you look in the HTML files,
+# you will see them accessed via {{ site.title }}, {{ site.github_repo }}, and so on.
+# You can create any custom variable you would like, and they will be accessible
+# in the templates via {{ site.myvariable }}.
+title: Konveyor MoveKube
+description: A Jekyll theme for documentation
```
ashokponkumar, line 17: Can you change the description to Move2Kube description?

_config.yml
```diff
@@ -1,21 +1,121 @@
...
+  - url: https://konveyor.github.io/
+    title: "&nbsp;&emsp;&emsp;Konveyor"
+  - url: https://groups.google.com/g/konveyorio
+    title: "&nbsp;&emsp;&emsp;Forum"
+  - url: https://github.com/konveyor/move2kube
+    title: "&nbsp;&emsp;&emsp;View on GitHub"
+
+# Google Analytics Tracking (optional)
+# e.g, UA-1234567-89
+ga_tracking: UA-2709176-10
```
ashokponkumar, line 109: Where did you get this id from? We should use the Konveyor repo google account

ashokponkumar (changes requested):

_config.yml
Akash-Nayak: Yes sure, I have changed it. It was there on my to-do-list but I missed correcting it.

_config.yml
Akash-Nayak: Oh, it came by default from the `just-the-docs` theme. Please suggest if I should I change it to `G-J08VDXBTDB` which is mentioned in the _config.yml of konveyor.github.io or should I comment this part for now?
Screenshot 2020-11-10 at 6 46 57 PM

ashokponkumar: Yes, use "G-J08VDXBTDB"

Akash-Nayak: Yes sure, I have updated it.

PR #211, opened 2020-11-13
print the service urls when the deploy script is run.

HarikrishnanBalagopal:


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	code.cloudfoundry.org/cli v7.1.0+incompatible
 	github.com/AlecAivazis/survey/v2 v2.1.1
 	github.com/Masterminds/semver/v3 v3.1.0
+	github.com/apex/log v1.3.0
```
ashokponkumar, line 10: We don't need it I assume?

ashokponkumar (changes requested):

go.mod
HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #223, opened 2020-11-15
fix bug where APPNAME was going out of scope and CMD [${APPNAME}] fails

HarikrishnanBalagopal:


Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN cp bin/${APPNAME} /bin/${APPNAME}
 
 # Run image
 FROM registry.access.redhat.com/ubi8/ubi:latest
```
ashokponkumar, line 40: If we are not using yum install, we can possibly change the image to minimal

HarikrishnanBalagopal: I think it requires more investigation.

PR #227, opened 2020-11-15
moved ReadPlan, WritePlan and SetRootDir out of internals. Made SetRootDir into a method.

HarikrishnanBalagopal:


types/plan/planutils.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	if err := convertPathsDecode(&plan); err != nil {
+		return plan, err
+	}
+	return plan, nil
+}
+
+// WritePlan encodes the plan to yaml converting absolute paths to relative.
+func WritePlan(path string, plan Plan) error {
```
ashokponkumar, line 200: Would it be better if it is a instance method too?

ashokponkumar (changes requested):

ashokponkumar (approved):

PR #241, opened 2020-11-20
add containers if images need to be built while translating docker-compose

HarikrishnanBalagopal:
previously we only allowed reuse of existing images. Now we support
building images using Dockerfiles mentioned in the compose yaml.

### DIFF
--- a/internal/source/compose/utils.go
+++ b/internal/source/compose/utils.go
@@ -20,8 +20,11 @@ import (
 	"fmt"
 	"hash/fnv"
 	"os"
+	"path/filepath"
 	"strings"
 
+	"github.com/konveyor/move2kube/internal/collector/sourcetypes"
+	"github.com/konveyor/move2kube/internal/common"
 	log "github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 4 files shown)


ashokponkumar (changes requested):
Dont' we want to add this support for V2 too?

internal/source/compose/v3.go
```diff
@@ -140,11 +157,22 @@ func (c *V3Loader) convertToIR(filedir string, composeObject types.Config, plan
...
+
+			// Context path is mandatory, Dockerfile is optional https://docs.docker.com/compose/compose-file/#build
+			composeFileDir := filedir
+			contextPath := filepath.Join(composeFileDir, composeServiceConfig.Build.Context)
+			dockerfilePath := filepath.Join(contextPath, "Dockerfile")
+			if composeServiceConfig.Build.Dockerfile != "" {
+				dockerfilePath = filepath.Join(contextPath, composeServiceConfig.Build.Dockerfile)
+			}
+			if checkForDockerfile(dockerfilePath) {
+				service.ContainerizationTargetOptions = append(service.ContainerizationTargetOptions, dockerfilePath)
```
ashokponkumar, line 169: We have to make it available as a containerisation option in plan file itself. Not just in the translate phase.

ashokponkumar (changes requested):

internal/source/compose/v3.go
HarikrishnanBalagopal: done

internal/source/compose/utils.go
```diff
@@ -31,6 +42,157 @@ const (
...
+	case "3", "3.0", "3.1", "3.2", "3.3", "3.4", "3.5", "3.6", "3.7", "3.8":
+		return true, nil
+	default:
+		err := fmt.Errorf("The compose file at path %s uses Docker Compose version %s which is not supported. Please use version 1, 2 or 3", path, dcfile.Version)
+		log.Errorf("Error: %q", err)
+		return false, err
+	}
+}
+
+// ParseV2 parses version 2 compose files
```
ashokponkumar, line 64: Would it not be better to move parseV2 and parseV3 functions to the respective version files instead of putting it in utils.go?

ashokponkumar (changes requested):

ashokponkumar (approved):
We can consider moving it as part of the loader class itself when we touch this file next.

PR #287, opened 2020-11-29
upgrade to openshift api 4.6 and fix domain name and path in routes

HarikrishnanBalagopal:

### DIFF
--- a/internal/apiresource/service.go
+++ b/internal/apiresource/service.go
@@ -18,7 +18,6 @@ package apiresource
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/konveyor/move2kube/internal/common"
 	irtypes "github.com/konveyor/move2kube/internal/types"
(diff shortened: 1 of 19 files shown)


internal/apiresource/service.go
```diff
@@ -387,17 +399,14 @@ func (d *Service) createRoute(service irtypes.Service) *okdroutev1.Route {
 			Labels: getServiceLabels(service.Name),
 		},
 		Spec: okdroutev1.RouteSpec{
-			Port: &okdroutev1.RoutePort{
-				TargetPort: intstr.IntOrString{
-					Type:   intstr.Int,
-					IntVal: common.DefaultServicePort,
-				},
-			},
+			Host: ir.TargetClusterSpec.Host,
```
ashokponkumar, line 402: Do we need to specify host? If it takes default, we can ignore it. Isn't it?

internal/apiresource/service.go
```diff
@@ -369,13 +371,23 @@ func (d *Service) ingressToService(ingress networkingv1.Ingress) []runtime.Objec
 	return objs
 }
 
+func (d *Service) createRoutes(service irtypes.Service, ir irtypes.IR) [](*okdroutev1.Route) {
```
ashokponkumar, line 374: Should we also fix ingressToRoute function and servicetoroute function? Also the routetoingress and other similar functions?

ashokponkumar (changes requested):

internal/apiresource/service.go
HarikrishnanBalagopal: the default is a subdomain that uses the route name

ashokponkumar (approved):

ashokponkumar (dismissed):

ashokponkumar (approved):

PR #305, opened 2020-12-04
add support for openshift build config

HarikrishnanBalagopal:
When targeting Openshift, generate a BuildConfig if Tekton
is not available for CI/CD purposes.


internal/apiresource/buildconfig.go
```diff
@@ -0,0 +1,184 @@
...
+func (*BuildConfig) getWebHookType(gitRepoURL string) okdbuildv1.BuildTriggerType {
+	switch true {
+	case strings.Contains(gitRepoURL, "github"):
+		return okdbuildv1.GitHubWebHookBuildTriggerType
+	case strings.Contains(gitRepoURL, "gitlab"):
+		return okdbuildv1.GitLabWebHookBuildTriggerType
+	case strings.Contains(gitRepoURL, "bitbucket"):
+		return okdbuildv1.BitbucketWebHookBuildTriggerType
+	default:
+		log.Warnf("the git repo is not on github, gitlab or bitbucket. Defaulting to generic web hook trigger")
```
ashokponkumar, line 170: Do we want this to be a debug message? Since in many cases the we might not have found a git repo url

ashokponkumar (changes requested):
Nice work Hari. Overall it looks great.

Few comments:

Do we have conditions to ensure that if a Openshift cluster (has buildconfig) supports tekton we create only tekton? I did not see any such condition in buildconfig createnewresource.

Also, do should we create buildconfig and imagestream even if the cluster supports only deployment (means somebody has manually disabled deployment config). This is a corner case, we can handle it as part of sensible defaults (we can create a new issue for it).

internal/apiresource/buildconfig.go
HarikrishnanBalagopal: ```Do we have conditions to ensure that if a Openshift cluster (has buildconfig) supports tekton we create only tekton? I did not see any such condition in buildconfig createnewresource.```
https://github.com/HarikrishnanBalagopal/move2kube/blob/buildconfig/internal/transformer/cicdtransformer.go#L39-L50

HarikrishnanBalagopal:
```
Also, do should we create buildconfig and imagestream even if the cluster supports only deployment (means somebody has manually disabled deployment config). This is a corner case, we can handle it as part of sensible defaults (we can create a new issue for it).
```
To get k8s deployments to use image streams we have to add certain annotations https://developers.redhat.com/blog/2019/09/20/using-red-hat-openshift-image-streams-with-kubernetes-deployments/
```
  annotations:
    deployment.kubernetes.io/revision: "2"
    image.openshift.io/triggers: '[{"from":{"kind":"ImageStreamTag","name":"ubi:8"},"fieldPath":"spec.template.spec.containers[?(@.name==\"myapp\")].image"}]'
```

HarikrishnanBalagopal: > 
> 
> Do we want this to be a debug message? Since in many cases the we might not have found a git repo url

If there's no repo should we just avoid creating CICD resources?

ashokponkumar: We can create it if there is request from customers. For now, create an issue with this link

ashokponkumar: > > 
> > 
> > Do we want this to be a debug message? Since in many cases the we might not have found a git repo url
> 
> If there's no repo should we just avoid creating CICD resources?

Let it be there, with placeholder for git repo.

ashokponkumar (dismissed):

ashokponkumar (approved):

ashokponkumar (dismissed):

ashokponkumar (approved):

PR #311, opened 2020-12-05
better error message when we don't find any services

### LINKED ISSUE #158: Better error message when `move2kube translate -s foo/` is run and no services are found.
When we do the one-step translation and it fails to find any services, it gives an unhelpful error message.
The current error message talks about being unable to fetch the answer.

Give a better error, something like: `Failed to find any services to translate.`

HarikrishnanBalagopal:

fix #158

### DIFF
--- a/internal/move2kube/planner.go
+++ b/internal/move2kube/planner.go
@@ -69,6 +69,9 @@ func CuratePlan(p plantypes.Plan) plantypes.Plan {
 		cachepaths = append(cachepaths, p.Spec.Inputs.QACaches[i])
 	}
 	qaengine.AddCaches(cachepaths)
+	if len(p.Spec.Inputs.Services) == 0 {
+		log.Fatalf("Failed to find any services. Aborting.")
+	}
 	// Identify services of interest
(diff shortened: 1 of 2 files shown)


internal/move2kube/planner.go
```diff
@@ -69,6 +69,9 @@ func CuratePlan(p plantypes.Plan) plantypes.Plan {
 		cachepaths = append(cachepaths, p.Spec.Inputs.QACaches[i])
 	}
 	qaengine.AddCaches(cachepaths)
+	if len(p.Spec.Inputs.Services) == 0 {
```
ashokponkumar, line 72: We have to do the same even if curate option is not selected. This code gets invoked only if we invoke the single command flow or translate -c.

ashokponkumar (changes requested):

internal/move2kube/planner.go
HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #312, opened 2020-12-06
added release-drafter configuration file

HarikrishnanBalagopal:

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,12 @@
+prerelease: true
+categories:
+  - title: "🚀 Features"
+    labels:
+      - "enhancement"
+  - title: "🐛 Bug Fixes"
+    labels:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,16 @@
+name: Release Drafter
```
ashokponkumar, line 1: We already have this as part of the release flow. Check release.yml

.github/release-drafter.yml
```diff
@@ -0,0 +1,4 @@
+template: |
+  # Changelog
+
+  $CHANGES
```
ashokponkumar, line 4: Do we want to specify any specific format or a labelling technique based format , so that major, minor, features are listed fine?

ashokponkumar (changes requested):

PR #324, opened 2020-12-11
fix: use an action to get the go version instead of hardcoding it

HarikrishnanBalagopal:


ashokponkumar (changes requested):
Can we add that chahe to include the docker file?

HarikrishnanBalagopal:
> Can we parameterize the version in dockerfile too?

done

ashokponkumar (approved):

PR #330, opened 2020-12-14
ci: added support for specifying commit and previous tag while releasing

HarikrishnanBalagopal:
For major releases we might want to have the changes from all
the pre-releases also included in the release notes.

We also might want to use a different commit to do the release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,16 +1,28 @@
-name: Release
+name: Tag and Release
 
-on: create
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "current tag: The tag for this release"
+        required: true
+        default: v0.1.0-rc.2
+      prev_tag:
+        description: "previous tag: Tag from which to start calculating the changelog"
+        required: false
+      commit_ref:
+        description: "commit ref: The branch, tag or SHA of the commit to use for the release."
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -70,27 +82,63 @@ jobs:
...
+          input_string: ${{ github.event.inputs.tag }}
       - uses: actions/setup-node@v1
         with:
           node-version: "12"
       - name: create release draft
         run: |
           npm install github-release-notes@v0.17.3 -g
-          gren release --draft --prerelease --tags ${{ steps.only_tag.outputs.tag }} --username=${{ github.repository_owner }} --repo=${{ github.event.repository.name }} --token=${{ secrets.GITHUB_TOKEN }}
+          if [[ '${{ github.event.inputs.prev_tag }}' ]] ; then
+            if [[ '${{ steps.semver_parser.outputs.prerelease }}' ]] ; then
```
ashokponkumar, line 130: Instead of copy pasting the entire command, would it be possible to have only the flag added to the string in case of pre-release, so that when we fix future bugs, we dpnt miss out on changing all commands.

ashokponkumar (changes requested):

ashokponkumar (approved):

PR #332, opened 2020-12-16
feat: create the translate plugin for kubectl

HarikrishnanBalagopal:
The translate plugin is simply the translate
command of move2kube.


Makefile
```diff
@@ -67,19 +67,25 @@ help: ## This help.
 # -- Build --
 
 .PHONY: build
-build: get $(BINDIR)/$(BINNAME) ## Build go code
+build: get $(BINDIR)/$(BINNAME) # Build go code
```
ashokponkumar, line 70: We need ## to show up as help, when just typing 'make'

Makefile
```diff
@@ -67,19 +67,25 @@ help: ## This help.
...
 
 $(BINDIR)/$(BINNAME): $(SRC)
 	go build -tags excludecodegen,excludedist -ldflags '$(LDFLAGS)' -o $(BINDIR)/$(BINNAME) ./cmd/${BINNAME}
 	mkdir -p $(GOPATH)/bin/
 	cp $(BINDIR)/$(BINNAME) $(GOPATH)/bin/
 
+.PHONY: build-translate
```
ashokponkumar, line 77: We need to run as dist with build-cross, so that binary is created for all platforms too. Do we want to name it as kubectl-translate, so that even in user's laptop it runs as a plugin.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+	qadisablecliFlag = "qadisablecli"
+	qaskipFlag       = "qaskip"
+	qaportFlag       = "qaport"
+	qacacheFlag      = "qacache"
+
+	nameFlag      = "name"
+	planFlag      = "plan"
+	sourceFlag    = "source"
+	ignoreEnvFlag = "ignoreenv"
+)
```
ashokponkumar, line 57: for the variables and values it shares with normal cmd, can we put it as a common one, so that they don't become inconsistent? Similarly for the functions. Only those functions/variables that we do change should be here. We can possible have some common files in cmd folder.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+	must := func(err error) {
+		if err != nil {
+			panic(err)
+		}
+	}
+	viper.AutomaticEnv()
+
+	flags := translateFlags{}
+	translateCmd := &cobra.Command{
+		Use:   "translate",
```
ashokponkumar, line 199: The help command should change depending on how it was invoked. We need to make that change in the move2kube binary too.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+		Short: "Translate using move2kube plan",
```
ashokponkumar, line 200: We need to change this. When users use plan, they are not using plan directly. Here we are targeting a different usecase of direct translation.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+			translateHandler(cmd, flags)
+		},
+	}
+
+	// Options inherited from move2kube root command
+	translateCmd.Flags().BoolVarP(&verbose, "verbose", "v", false, "Enable verbose output")
+
+	// Basic options
+	translateCmd.Flags().StringVarP(&flags.planfile, planFlag, "p", common.DefaultPlanFile, "Specify a plan file to execute.")
+	translateCmd.Flags().BoolVarP(&flags.curate, curateFlag, "c", false, "Specify whether to curate the plan with a q/a.")
```
ashokponkumar, line 215: We need to decide whether to talk a lot of plan in the standalone command.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+	translateCmd.Flags().StringVarP(&flags.srcpath, sourceFlag, "s", "", "Specify source directory to translate. If you already have a m2k.plan then this will override the rootdir value specified in that plan.")
+	translateCmd.Flags().StringVarP(&flags.outpath, outpathFlag, "o", ".", "Path for output. Default will be directory with the project name.")
+	translateCmd.Flags().StringVarP(&flags.name, nameFlag, "n", common.DefaultProjectName, "Specify the project name.")
+	translateCmd.Flags().StringSliceVarP(&flags.qacaches, qacacheFlag, "q", []string{}, "Specify qa cache file locations")
+
+	// Advanced options
+	translateCmd.Flags().BoolVar(&flags.ignoreEnv, ignoreEnvFlag, false, "Ignore data from local machine.")
+
+	// Hidden options
+	translateCmd.Flags().BoolVar(&flags.qadisablecli, qadisablecliFlag, false, "Enable/disable the QA Cli sub-system. Without this system, you will have to use the REST API to interact.")
```
ashokponkumar, line 225: We need not support disabling qa cli.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+	translateCmd.Flags().BoolVar(&flags.qaskip, qaskipFlag, false, "Enable/disable the default answers to questions posed in QA Cli sub-system. If disabled, you will have to answer the questions posed by QA during interaction.")
+	translateCmd.Flags().IntVar(&flags.qaport, qaportFlag, 0, "Port for the QA service. By default it chooses a random free port.")
```
ashokponkumar, line 227: If we don't support disabling cli, this becomes redundant.

cmd/translate/translate.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		checkSourcePath(srcpath)
+		log.Debugf("Creating a new plan.")
+		p = move2kube.CreatePlan(srcpath, name)
+		outpath = filepath.Join(outpath, p.Name)
+		checkOutputPath(outpath)
+		createOutputDirectoryAndCacheFile(outpath)
+		p = move2kube.CuratePlan(p)
+	} else {
+		log.Infof("Detected a plan file at path %s. Will translate using this plan.", planfile)
```
ashokponkumar, line 159: I was thinking we can have a simple flow for just the translate command which avoids the plan file. Makes it a simple flow. Even though users know about it (just because it is in their system), the command itself (kubectl translate), has no ref to move2kube, so we can go with more streamlined flow here.

ashokponkumar (changes requested):

Makefile
HarikrishnanBalagopal: I was wondering why we had that. done

Makefile
HarikrishnanBalagopal: done.

cmd/translate/translate.go
HarikrishnanBalagopal: you can't import from `main` package so it will have to be a separate folder and package called `common`

cmd/translate/translate.go
HarikrishnanBalagopal: what does it need to say?

cmd/translate/translate.go
HarikrishnanBalagopal: it's a good place to advertise move2kube

HarikrishnanBalagopal: we can link to the main repo and `konveyor.io/move2kube` in the help.
Mention that there's an even more powerful/flexible version in the main repo.

Makefile
ashokponkumar: yes, build command should put it in path. line 75 does it for example

cmd/translate/translate.go
ashokponkumar: it need not be named main.

cmd/translate/translate.go
ashokponkumar: the command should be either "kubectl-translate" or "kubectl translate". Checkout "Revise usage/help messages with kubectl prefix" section in https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/

cmd/common/utils.go
```diff
@@ -0,0 +1,85 @@
...
+	QacacheFlag = "qacache"
+	// NameFlag is the name of the flag that contains the project name
+	NameFlag = "name"
+	// PlanFlag is the name of the flag that contains the path to the plan file.
+	PlanFlag = "plan"
+	// IgnoreEnvFlag is the name of the flag that tells us whether to use data collected from the local machine.
+	IgnoreEnvFlag = "ignoreenv"
+)
+
+var verbose bool
```
ashokponkumar, line 43: Are we using this verbose flag?

cmd/kubectltranslate/kubectltranslate.go
```diff
@@ -0,0 +1,134 @@
...
+	// Setup
+	var err error
+	ignoreEnv := false
+	srcpath := flags.srcpath
+	outpath := flags.outpath
+	name := flags.name
+
+	qadisablecli := false
+	qaskip := false
+	qaport := 0
```
ashokponkumar, line 51: Can we add a comment here why we are setting it to 0?

cmd/kubectltranslate/kubectltranslate.go
```diff
@@ -0,0 +1,134 @@
...
+	name     string
+	qacaches []string
+}
+
+var verbose bool
+
+func translateHandler(cmd *cobra.Command, flags translateFlags) {
+	// Setup
+	var err error
+	ignoreEnv := false
```
ashokponkumar, line 44: Let's add a comment why we are setting ignoreenv as false

cmd/kubectltranslate/kubectltranslate.go
```diff
@@ -0,0 +1,134 @@
...
+	srcpath := flags.srcpath
+	outpath := flags.outpath
+	name := flags.name
+
+	qadisablecli := false
```
ashokponkumar, line 49: Let's add a comment why we are setting qadisablecli as false

cmd/kubectltranslate/kubectltranslate.go
```diff
@@ -0,0 +1,134 @@
...
+	qaskip := false
```
ashokponkumar, line 50: qaskip should be allowed in translate. It should not be hardcoded

cmd/kubectltranslate/kubectltranslate.go
```diff
@@ -0,0 +1,134 @@
...
+	must := func(err error) {
+		if err != nil {
+			panic(err)
+		}
+	}
+	viper.AutomaticEnv()
+
+	flags := translateFlags{}
+	translateCmd := &cobra.Command{
+		Use:   "translate [ -o <output directory> ] [ -n <project name> ] [ -q <list of qacache files> ] -s <source directory>",
```
ashokponkumar, line 94: The command should update based on usage. https://krew.sigs.k8s.io/docs/developer-guide/develop/best-practices/

ashokponkumar (changes requested):

cmd/common/utils.go
HarikrishnanBalagopal: oops. removed.

cmd/kubectltranslate/kubectltranslate.go
HarikrishnanBalagopal: because that's the default https://github.com/konveyor/move2kube/blob/main/cmd/move2kube/translate.go#L211

ashokponkumar: I meant add a comment in code saying the reason. Something like:

 // setting 0, since kubectl-translate does not support api based access.

cmd/kubectltranslate/kubectltranslate.go
HarikrishnanBalagopal: done

cmd/kubectltranslate/kubectltranslate.go
HarikrishnanBalagopal: done

cmd/kubectltranslate/kubectltranslate.go
HarikrishnanBalagopal: done

cmd/kubectltranslate/kubectltranslate.go
HarikrishnanBalagopal: what?

ashokponkumar (approved):

ashokponkumar (approved):

PR #341, opened 2020-12-19
chore: workflow changes to accomodate the new release branches

HarikrishnanBalagopal:

### DIFF
--- a/.github/workflows/prbuild.yml
+++ b/.github/workflows/prbuild.yml
@@ -2,7 +2,9 @@ name: PR Build
 
 on:
   pull_request:
-    branches: [main]
+    branches:
+      - "main"
+      - "release-*"
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,9 @@ name: Build
 
 on:
   push:
-    branches: [main]
+    branches:
+      - "main"
+      - "release-*"
 
 jobs:
   build_and_test:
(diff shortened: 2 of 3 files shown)


.github/workflows/prbuild.yml
```diff
@@ -17,5 +19,13 @@ jobs:
...
       - run: make test-coverage
-      - name: upload coverage
+      - id: branch_info
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const resp = await github.request('GET /repos/{owner}/{repo}/pulls/{pull_number}', {...context.repo, pull_number:context.payload.number});
+            core.setOutput('branch', resp.data.base.ref);
+      - if: ${{ steps.branch_info.outputs.branch == 'main' }}
```
ashokponkumar, line 29: Why not upload coverage for all branches?

ashokponkumar (changes requested):

.github/workflows/prbuild.yml
HarikrishnanBalagopal: Not sure if that's how it works.
Doesn't it just overwrite the coverage of the main branch?

ashokponkumar: I don't think so. The ui seems to have branches and stuff. Even if not the action would take as input.

HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #343, opened 2020-12-21
revert: Delta builds not working with make build

ashokponkumar:


ashokponkumar:
@HarikrishnanBalagopal The recent change to print0 in Makefile is breaking the delta builds using `make build`. Basically print0 prints without a newline between files, and that breaks the delta check. Was there a reason that compelled us to change to print0?

HarikrishnanBalagopal (approved):

PR #354, opened 2020-12-25
feat: write release information to a file

HarikrishnanBalagopal:
This info will be displayed on the website.
This info will also be used by the release
UI to decide the tag, commit, etc. to use
for new releases.

Also ignore node_modules in main branch to
allow easy switching to and from gh-pages.

### DIFF
--- a/.github/workflows/releaseinfo.yml
+++ b/.github/workflows/releaseinfo.yml
@@ -1,0 +1,26 @@
+name: write release information to file
+
+on: workflow_dispatch
+
+jobs:
+  release_info:
+    runs-on: ubuntu-latest
+    steps:
+      - id: info
+        uses: HarikrishnanBalagopal/nextrelease@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+      - run: echo '${{ steps.info.outputs.release_info }}' > releaseinfo.json
+      - uses: EndBug/add-and-commit@v6
+        with:
+          add: releaseinfo.json
+          author_name: move2kube
+          author_email: move2kube@gmail.com
+          branch: gh-pages
+          cwd: "."
+          message: "chore: update release information. run_id ${{ github.run_id }}"
+          pull_strategy: "--ff-only"
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/releaseinfo.yml
+++ b/.github/workflows/releaseinfo.yml
@@ -7,16 +7,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - id: info
-        uses: HarikrishnanBalagopal/nextrelease@v1
+        uses: konveyor/get-release-info@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
       - uses: actions/checkout@v2
         with:
           ref: gh-pages
-      - run: echo '${{ steps.info.outputs.release_info }}' > releaseinfo.json
+      - run: echo '${{ steps.info.outputs.release_info }}' > _data/releaseinfo.json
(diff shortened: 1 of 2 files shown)


.github/workflows/releaseinfo.yml
```diff
@@ -0,0 +1,26 @@
+name: write release information to file
+
+on: workflow_dispatch
+
+jobs:
+  release_info:
+    runs-on: ubuntu-latest
+    steps:
+      - id: info
+        uses: HarikrishnanBalagopal/nextrelease@v1
```
ashokponkumar, line 10: This repo should be in Konveyor org. Can we name it move2kube-nextrelease?

.github/workflows/releaseinfo.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+      - run: echo '${{ steps.info.outputs.release_info }}' > _data/releaseinfo.json
```
ashokponkumar, line 16: Do we need the "_"? The data in it is similar to bin, scripts and assets, rather than like _include,..

.github/workflows/releaseinfo.yml
```diff
@@ -0,0 +1,26 @@
+name: write release information to file
+
+on: workflow_dispatch
```
ashokponkumar, line 3: Does it need to be a separate workflow? Why not make it as part of "publish" workflow?

ashokponkumar (changes requested):

.github/workflows/releaseinfo.yml
HarikrishnanBalagopal: yes _data is for data files. https://jekyllrb.com/docs/datafiles/

.github/workflows/releaseinfo.yml
HarikrishnanBalagopal: Doesn't that require a lot of approvals?

ashokponkumar: No. Only if we need to use generic names or temp repo we need to ensure there is a valid reason.

.github/workflows/releaseinfo.yml
ashokponkumar: Okay cool

.github/workflows/releaseinfo.yml
HarikrishnanBalagopal: I think it's best to keep the workflows simple.
It's already getting very complicated.

HarikrishnanBalagopal: We also need to write `index.yaml` for helm charts, so this can be repurposed for that even.

ashokponkumar: Okay. Who will call this workflow?

.github/workflows/releaseinfo.yml
HarikrishnanBalagopal: So can you create the repo? I will upload the code there.

.github/workflows/releaseinfo.yml
HarikrishnanBalagopal: The publish workflow will call this.

.github/workflows/releaseinfo.yml
ashokponkumar: Sure, Name? move2kube-nextreleaseinfo-action?

.github/workflows/releaseinfo.yml
ashokponkumar: Okay cool.

HarikrishnanBalagopal: also keeping it separate means we can trigger it manually if we want.

ashokponkumar (approved):

PR #356, opened 2020-12-27
fix: show error when credentials are wrong

HarikrishnanBalagopal:


assets/ts/internalpages/releaseui.ts
```diff
@@ -310,7 +310,11 @@ async function setup(): Promise<void> {
             return alert(`personal access token is invalid. token should only contain hexadecimal characters [a-f0-9]. got: ${token}`);
```
ashokponkumar, line 310: Do we want to display the pa token as part of the message? Isn't it a sort of secret? Also, would be good to put the first letter 'P' and 'T'  as capital.

ashokponkumar (changes requested):

assets/ts/internalpages/releaseui.ts
HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #363, opened 2020-12-28
docs: Add doc for all-in-one container bringup

ashokponkumar:

### DIFF
--- a/docs/installation/web-interface.md
+++ b/docs/installation/web-interface.md
@@ -8,11 +8,23 @@ nav_order: 2
 
 # Move2Kube Web Interface
 
+## Bringing up Move2Kube all-in-one container
+
+   ```console
+   $ mkdir -p workspace && cd workspace
+   $ docker run -p 8080:8080 -v $PWD:/workspace -v /var/run/docker.sock:/var/run/docker.sock -it quay.io/konveyor/move2kube-aio:latest
+   ```
+   Access the UI in `http://localhost:8080/`.
+
+## Bringing up Move2Kube UI and API as separate containers
+
    ```console
    $ git clone https://github.com/konveyor/move2kube-ui
    $ cd move2kube-ui
    $ docker-compose up
    ```
+   Access the UI in `http://localhost:8080/`.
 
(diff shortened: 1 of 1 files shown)


docs/installation/web-interface.md
```diff
@@ -8,11 +8,23 @@ nav_order: 2
 
 # Move2Kube Web Interface
 
+## Bringing up Move2Kube all-in-one container
+
+   ```console
+   $ mkdir -p workspace && cd workspace
+   $ docker run -p 8080:8080 -v $PWD:/workspace -v /var/run/docker.sock:/var/run/docker.sock -it quay.io/konveyor/move2kube-aio:latest
```
Akash-Nayak, line 15: I am able to run the all-in-one container successfully and access the UI at `http://localhost:8080/` using these two commands. Then, I tried translating the cloud foundry sample usecase. After the translation was completed, when I tried to download the generated target artifacts, it downloaded a single json file. Tried doing the translation for the language platforms sample and for that also a json file was downloaded when I clicked on `Get`.

Screenshot 2020-12-28 at 7 49 37 PM

ashokponkumar: Can you try with chrome once? And also in the current one can you check the url pointed to by get?

Akash-Nayak: Yes, I tried with chrome it is working fine and I am able to download the zip file containing the generated artifacts. Hovering over the `Get` does not show the url, in the browser console the filename is `/api/v1/applications/newapp/targetartifacts/newapp_1609164843`.

Screenshot 2020-12-28 at 8 15 16 PM

ashokponkumar: Then it's some firefox compatibility issue. Let's merge this and create a new issue in move2kube-ui? Btw, did the docker-compose method work in Firefox?

Akash-Nayak: Yes sure. I just checked with docker-compose method also, I am facing the same issue with that as well.

Akash-Nayak (approved):

Issue #155, opened 2020-11-03
Enhancement: Consider adding some tests that run periodically to ensure the html processing is not broken.


HarikrishnanBalagopal:
Possibly use IBM cloud functions to run periodic tests for this:
https://github.com/konveyor/move2kube/blob/b68649ed6a2217e9990bcc0b7828fd9c761fdd68/scripts/install.sh#L84-L86

PR #378, opened 2021-01-04
chore: use releaseinfo.json to get the latest stable release

HarikrishnanBalagopal:

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -24,6 +24,7 @@
 [[ $VERIFY_CHECKSUM ]] || VERIFY_CHECKSUM='true'
 [[ $MOVE2KUBE_INSTALL_DIR ]] || MOVE2KUBE_INSTALL_DIR='/usr/local/bin'
 
+HAS_JQ="$(type jq &>/dev/null && echo true || echo false)"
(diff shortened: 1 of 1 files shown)


HarikrishnanBalagopal:
Partial fix for https://github.com/konveyor/move2kube/issues/155

scripts/install.sh
```diff
@@ -78,12 +79,18 @@ verifySupported() {
 
 # getLatestVersion gets the latest release version.
 getLatestVersion() {
-    # Get tag from release URL
-    local latest_release_url='https://github.com/konveyor/move2kube/releases'
+    # Get tag from releaseinfo.json
+    local json_data=''
+    local release_info_url='https://raw.githubusercontent.com/konveyor/move2kube/gh-pages/_data/releaseinfo.json'
```
ashokponkumar, line 84: Is there no way to access the url using konveyor.io url as github pages?

ashokponkumar (commented):
Also, we wanted to install konveyor_translate binary too. Do we want to install it as part of this PR or a new PR?

scripts/install.sh
HarikrishnanBalagopal: why is that important?

ashokponkumar (approved):

Issue #137, opened 2020-10-29
Consider enhancing the qacache file to become a config file


HarikrishnanBalagopal:
# Config file format
More advanced users might want to configure move2kube.
Need to consider fixing the format of the qacache file so it can become a config file.
The format also needs to be clearly documented so people can use it.

# Config filename convention
Also consider enforcing a filename convention for the config files something like `m2kconfig-*.yaml`
Currently we look at any and all yaml files and try to parse them into the qacache struct.
https://github.com/konveyor/move2kube/blob/6fb351955ab146cc1a652d271d3c34022d05b097/internal/metadata/qacaches.go#L34-L41
This is very hard to debug. If a problem occurs (a service is not getting exposed, etc.) a user would have to go through every single yaml file to find out which ones are valid config files.

Most software already have a fixed filename or some filename format for their config files, so users will expect this when it comes to config files.

As a bonus it also allows you to easily enable/disable a config by simply renaming it. No need to delete it or move it to a different location.

PR #391, opened 2021-01-22
feat: Add support for config files

ashokponkumar:


HarikrishnanBalagopal:
fixes #137

HarikrishnanBalagopal:
@ashokponkumar can you take a look?

ashokponkumar:
Thanks for the update.

Few doubts:
 Don't we need to add writecache too to the list of readcaches?
 With yaml, do we still need CleanForIDSegment? 
 Why do we need a special string to indicate multiselect problems? From a config perspective our users need not be aware that qa and config are even the same engine. Isn't it?
 Are we merging maps? What happens to arrays? Are we merging them?

HarikrishnanBalagopal (approved):

PR #401, opened 2021-01-30
feat: Remove support for Kube2Kube and Knative2Kube as services

ashokponkumar:


HarikrishnanBalagopal (dismissed):
What about the warnings when run on `move2kube-demos/samples/kubernetes-to-kubernetes`?

HarikrishnanBalagopal (approved):

PR #403, opened 2021-02-01
feat: compress binaries

HarikrishnanBalagopal:


ashokponkumar:
Do we not want to install upx in our build system?

HarikrishnanBalagopal:
> Do we not want to install upx in our build system?

hmm, not really since it's not golang related

ashokponkumar:
> m, not really since it's not golang relate

I mean in github actions

ashokponkumar (approved):

PR #404, opened 2021-02-01
feat: Move fixers to a package and simplify convert logic

ashokponkumar:
feat: Move fixers to a package and simplify convert logic


internal/apiresource/apiresource.go
```diff
@@ -38,17 +39,16 @@ const (
 
 // IAPIResource defines the interface to be defined for a new api resource
 type IAPIResource interface {
-	GetSupportedKinds() []string
-	CreateNewResources(ir irtypes.EnhancedIR, supportedKinds []string) []runtime.Object
+	getSupportedKinds() []string
```
HarikrishnanBalagopal, line 42: Why make the interface methods private?
We can make the interface private instead.

internal/apiresource/rolebinding.go
```diff
@@ -67,13 +67,7 @@ func (*RoleBinding) createNewResource(irrolebinding irtypes.RoleBinding) *rbac.R
...
-func (rb *RoleBinding) ConvertToClusterSupportedKinds(obj runtime.Object, supportedKinds []string, otherobjs []runtime.Object, _ irtypes.EnhancedIR) ([]runtime.Object, bool) {
-	supKinds := rb.GetSupportedKinds()
-	for _, supKind := range supKinds {
-		if common.IsStringPresent(supportedKinds, supKind) {
-			return []runtime.Object{obj}, true
-		}
-	}
-	return nil, false
+// convertToClusterSupportedKinds converts the object to supported types if possible.
+func (rb *RoleBinding) convertToClusterSupportedKinds(obj runtime.Object, supportedKinds []string, otherobjs []runtime.Object, _ irtypes.EnhancedIR) ([]runtime.Object, bool) {
```
HarikrishnanBalagopal, line 71: why are these checks removed?

internal/apiresource/triggerbinding.go
```diff
@@ -57,13 +56,7 @@ func (*TriggerBinding) createNewResource(irtriggerbinding tekton.TriggerBinding)
...
-	supKinds := tb.GetSupportedKinds()
-	for _, supKind := range supKinds {
-		if common.IsStringPresent(supportedKinds, supKind) {
-			return []runtime.Object{obj}, true
-		}
-	}
-	return nil, false
+// convertToClusterSupportedKinds converts the object to supported types if possible.
+func (tb *TriggerBinding) convertToClusterSupportedKinds(obj runtime.Object, supportedKinds []string, otherobjs []runtime.Object, _ irtypes.EnhancedIR) ([]runtime.Object, bool) {
+	return []runtime.Object{obj}, true
```
HarikrishnanBalagopal, line 61: same as above.

internal/apiresource/triggertemplate.go
```diff
@@ -112,13 +112,7 @@ func (*TriggerTemplate) createNewResource(tt tekton.TriggerTemplate, ir irtypes.
...
-func (tt *TriggerTemplate) ConvertToClusterSupportedKinds(obj runtime.Object, supportedKinds []string, otherobjs []runtime.Object, _ irtypes.EnhancedIR) ([]runtime.Object, bool) {
-	supKinds := tt.GetSupportedKinds()
-	for _, supKind := range supKinds {
-		if common.IsStringPresent(supportedKinds, supKind) {
-			return []runtime.Object{obj}, true
-		}
-	}
-	return nil, false
+// convertToClusterSupportedKinds converts the object to supported types if possible.
+func (tt *TriggerTemplate) convertToClusterSupportedKinds(obj runtime.Object, supportedKinds []string, otherobjs []runtime.Object, _ irtypes.EnhancedIR) ([]runtime.Object, bool) {
```
HarikrishnanBalagopal, line 116: same as above.

internal/apiresource/apiresource.go
ashokponkumar: The interface is still used by transformer which is outside this package.

internal/apiresource/rolebinding.go
ashokponkumar: this function is to moving between kind concepts if there is a similar concept supported in the cluster. In case there is no equivalent concept and the resource is required, it is better to retain the object. It is for the final writer to decide whether to serialize it or not.

internal/apiresource/triggerbinding.go
ashokponkumar: this function is to moving between kind concepts if there is a similar concept supported in the cluster. In case there is no equivalent concept and the resource is required, it is better to retain the object. It is for the final writer to decide whether to serialize it or not.

internal/apiresource/triggertemplate.go
ashokponkumar: this function is to moving between kind concepts if there is a similar concept supported in the cluster. In case there is no equivalent concept and the resource is required, it is better to retain the object. It is for the final writer to decide whether to serialize it or not.

HarikrishnanBalagopal (dismissed):
After adding checks in `convertToClusterSupportedKinds` we can merge

HarikrishnanBalagopal (approved):

PR #410, opened 2021-02-03
feat: source and output paths should not be parent directories of the current working directory

HarikrishnanBalagopal:

### DIFF
--- a/cmd/common/utils.go
+++ b/cmd/common/utils.go
@@ -19,6 +19,7 @@ package common
 import (
 	"os"
 
+	internalcommon "github.com/konveyor/move2kube/internal/common"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/common/utils.go
```diff
@@ -85,6 +86,13 @@ func CheckSourcePath(srcpath string) {
 	if !fi.IsDir() {
 		log.Fatalf("The given source path %s is a file. Expected a directory. Exiting.", srcpath)
 	}
+	pwd, err := os.Getwd()
```
ashokponkumar, line 89: Why are we concerned about the current working directory at all? Shouldn't we be just comparing between the source and output directories?

ashokponkumar (changes requested):

cmd/common/utils.go
HarikrishnanBalagopal: to check if the source is a parent of the current working directory.

HarikrishnanBalagopal:
```
we be just comparing between the source and output directories?
```
we are doing that too

ashokponkumar (approved):

PR #411, opened 2021-02-03
fix: arguments to IsParent were switched

HarikrishnanBalagopal:
also add the check to kubectl-translate plugin

### DIFF
--- a/cmd/kubectltranslate/kubectltranslate.go
+++ b/cmd/kubectltranslate/kubectltranslate.go
@@ -52,6 +52,9 @@ func translateHandler(cmd *cobra.Command, flags cmdcommon.TranslateFlags) {
 	cmdcommon.CheckSourcePath(flags.Srcpath)
 	flags.Outpath = filepath.Join(flags.Outpath, flags.Name)
 	cmdcommon.CheckOutputPath(flags.Outpath, flags.Overwrite)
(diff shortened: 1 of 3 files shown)


cmd/kubectltranslate/kubectltranslate.go
```diff
@@ -52,6 +52,9 @@ func translateHandler(cmd *cobra.Command, flags cmdcommon.TranslateFlags) {
 	cmdcommon.CheckSourcePath(flags.Srcpath)
 	flags.Outpath = filepath.Join(flags.Outpath, flags.Name)
 	cmdcommon.CheckOutputPath(flags.Outpath, flags.Overwrite)
+	if flags.Srcpath == flags.Outpath || common.IsParent(flags.Outpath, flags.Srcpath) {
```
ashokponkumar, line 55: Do we want to check the other way around too?

ashokponkumar (changes requested):

cmd/kubectltranslate/kubectltranslate.go
HarikrishnanBalagopal: `move2kube translate -s foo/bar/ -o foo/` should be fine

ashokponkumar: Exactly, we should be checking the final output folder, along with project name

ashokponkumar (approved):

PR #416, opened 2021-02-05
fix: parsing of config strings

HarikrishnanBalagopal:


ashokponkumar:
Why are we moving back from StringSlice? I thought StringSlice will provide more flexibility in which users can specify the flag as many times as they want.

HarikrishnanBalagopal:
> Why are we moving back from StringSlice? I thought StringSlice will provide more flexibility in which users can specify the flag as many times as they want.

It doesn't like double quotes https://github.com/open-policy-agent/opa/issues/2196

ashokponkumar (dismissed):

ashokponkumar (approved):

PR #418, opened 2021-02-06
chore: combine all 3 steps of asking a question into a single function call

HarikrishnanBalagopal:

### DIFF
--- a/internal/qaengine/engine.go
+++ b/internal/qaengine/engine.go
@@ -175,3 +175,90 @@ func changeSelectToInputForOther(prob qatypes.Problem) qatypes.Problem {
 	}
 	return prob
 }
+
+// Convenience functions
+
+// FetchStringAnswer asks a input type question and gets a string as the answer
+func FetchStringAnswer(probid, desc string, context []string, def string) string {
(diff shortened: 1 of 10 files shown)


internal/qaengine/engine.go
```diff
@@ -175,3 +175,90 @@ func changeSelectToInputForOther(prob qatypes.Problem) qatypes.Problem {
 	}
 	return prob
 }
+
+// Convenience functions
+
+// FetchStringAnswer asks a input type question and gets a string as the answer
```
ashokponkumar, line 181: Let's add one for multi line too. We don't use it now, but when users want it, they need not hunt and find out how to add it.

ashokponkumar (changes requested):

internal/qaengine/engine.go
HarikrishnanBalagopal: Actually I think we should remove MultiLine. Doesn't seem that useful.

ashokponkumar (approved):

internal/qaengine/engine.go
HarikrishnanBalagopal: done

PR #423, opened 2021-02-09
feat: Add validate framework to list TODO items after move2kube execution

ashokponkumar:


internal/move2kube/validate.go
```diff
@@ -0,0 +1,66 @@
...
+			log.Debugf("Failed to split the file at path %q into YAML documents. Error: %q", filePath, err)
+			continue
+		}
+		for _, doc := range docs {
+			obj, _, err := codecs.UniversalDeserializer().Decode(doc, nil, nil)
+			if err != nil {
+				continue
+			}
+			objectMeta := reflect.ValueOf(obj).Elem().FieldByName("ObjectMeta").Interface().(metav1.ObjectMeta)
+			for k, v := range objectMeta.Annotations {
```
HarikrishnanBalagopal, line 58: Why are we only checking the annotations?
The `TODO` placeholders can appear anywhere. Most of them are in the `spec`

HarikrishnanBalagopal (changes requested):

internal/move2kube/validate.go
ashokponkumar: We don't want it to appear anywhere else at least in Kubernetes yamls. The actual fields should have some defaults or empty string. We will have to move all those which are in spec as annotations.

cmd/move2kube/validate.go
```diff
@@ -0,0 +1,56 @@
...
+			panic(err)
+		}
+	}
+	viper.AutomaticEnv()
+
+	flags := validateFlags{}
+	validateCmd := &cobra.Command{
+		Use:   "validate",
+		Short: "Print all next steps in a directory",
+		Long:  "Print all next steps in a directory",
```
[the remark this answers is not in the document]
ashokponkumar: Have updated with a bit more information, we can do an exercise to update all the long messages together.

internal/move2kube/validate.go
ashokponkumar: As discussed, the actual comments will be stored in annotations for kubernetes yamls, and while writing to files, we can use different mechanisms to write them in comments and in other places.

HarikrishnanBalagopal (approved):

PR #424, opened 2021-02-09
chore: allow install script to install any tag

HarikrishnanBalagopal:


ashokponkumar (changes requested):
Would it be better to have the variable name as say `MOVE2KUBE_TAG`, and if it is set use the tag, if not, get the latest? Instead of having to specify 2 environment variables to use make it install a tag?

ashokponkumar (approved):

PR #434, opened 2021-02-15
fix: Reorder versions of the same kind to prioritize k8s versions

ashokponkumar:


internal/metadata/clusters/ibm-openshift.yaml
```diff
@@ -213,38 +213,38 @@ spec:
...
       - authorization.openshift.io/v1
     Route:
       - route.openshift.io/v1
     Scale:
-      - apps.openshift.io/v1
       - apps/v1
       - apps/v1beta1
       - apps/v1beta2
       - extensions/v1beta1
       - v1
```
HarikrishnanBalagopal, line 232: shouldn't v1 be at the top?

HarikrishnanBalagopal (changes requested):
Please fix https://github.com/konveyor/move2kube/blob/main/internal/metadata/clusters/gcp-gke.yaml#L72-L80

internal/metadata/clusters/openshift.yaml
```diff
@@ -194,24 +194,24 @@ spec:
...
       - authorization.openshift.io/v1
     Route:
       - route.openshift.io/v1
     Scale:
-      - apps.openshift.io/v1
       - apps/v1
       - apps/v1beta1
       - apps/v1beta2
       - extensions/v1beta1
       - v1
```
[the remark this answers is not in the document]
ashokponkumar: No. That v1 is from before groups concept was started.

internal/metadata/clusters/ibm-openshift.yaml
ashokponkumar: No. That v1 is from before groups concept was started.

HarikrishnanBalagopal (approved):

PR #436, opened 2021-02-16
fix: target path should be absolute

HarikrishnanBalagopal:


ashokponkumar (changes requested):
Do we need the go.mod and go.sum changes too?

ashokponkumar (approved):

PR #439, opened 2021-03-01
feat: add starlark scripting support

HarikrishnanBalagopal:


go.mod
```diff
@@ -33,8 +34,10 @@ require (
 	github.com/tektoncd/triggers v0.10.0
 	github.com/whilp/git-urls v1.0.0
 	github.com/xrash/smetrics v0.0.0-20200730060457-89a2a8a1fb0b
+	go.starlark.net v0.0.0-20210223155950-e043a3d3c984
 	golang.org/x/crypto v0.0.0-20201124201722-c8d3bf9c5392
 	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
+	gopkg.in/yaml.v2 v2.3.0
```
ashokponkumar, line 40: Can we change it to v3?

ashokponkumar (changes requested):

ashokponkumar (approved):

go.mod
HarikrishnanBalagopal: we need v2 to avoid some unmarshalling issues.
v3 unmarshals timestamps into `time.Time` objects.
v2 unmarshals timestamps into strings.
Can't marshal `time.Time` objects into `starlark.Value` since it is a struct type.

PR #440, opened 2021-03-03
Update _config.yml

iranzo:
Fix path for files


ashokponkumar:
Thanks for the fix @iranzo ! :)

PR #444, opened 2021-03-08
fix: javamaven dockerfile template: change to app folder and image needed

alealavi:
app folder was not being created, image was changed for openjdk 1.8 and works, only needed is change the .app_name to be the same as the current project name being translated

ex:
COPY --from=build_base /{{ .app_name }}/(app name missing here)/target/*.jar /deployments/

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,19 @@ Move2Kube is a command-line tool that accelerates the process of re-platforming 
 
 ### Using install script
 
+To install the latest stable version:
 ```
 bash <(curl https://raw.githubusercontent.com/konveyor/move2kube/main/scripts/install.sh)
+```
+
+To install a specific version (for example version `v0.2.0-alpha.3`):
+```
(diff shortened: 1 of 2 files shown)


HarikrishnanBalagopal (changes requested):
Please remake the PR from a branch other than `main`.
image

Example:
```
git checkout -b myfeature
git push -u origin myfeature
```
and then make the PR from the `myfeature` branch.

Also please sign the DCO.

alealavi:
Yes after the COPY, WORKDIR gives an error that cant access app, then I checked and app wasnt even created, could it be because of what you said on set as 755? it sounds like it shouldnt be a problem but making the directory as root solved that part.

And sure ill remake another PR with another branch from my fork

HarikrishnanBalagopal:
> Yes after the COPY, WORKDIR gives an error that cant access app, then I checked and app wasnt even created, could it be because of what you said on set as 755? it sounds like it shouldnt be a problem but making the directory as root solved that part.

Hmm, can you provide the input that gave the error and the error messages? Would like to reproduce the bug if possible.

> And sure ill remake another PR with another branch from my fork

Thanks!

PR #448, opened 2021-03-09
chore: allow installing the bleeding edge version

HarikrishnanBalagopal:


README.md
```diff
@@ -19,6 +19,16 @@ Move2Kube is a command-line tool that accelerates the process of re-platforming
...
 ```
 
+To install a specific version (for example version `v0.2.0-alpha.3`):
+```
+MOVE2KUBE_TAG='v0.2.0-alpha.3' bash <(curl https://raw.githubusercontent.com/konveyor/move2kube/main/scripts/install.sh)
+```
+
+To install the bleeding edge version:
+```
+BLEEDING_EDGE='true' bash <(curl https://raw.githubusercontent.com/konveyor/move2kube/main/scripts/install.sh)
```
ashokponkumar, line 29: Can we possibly call it say `INCLUDE_PRERELEASE` or something which includes the word pre-release to be explicity?

HarikrishnanBalagopal: it might not be prerelease

ashokponkumar: yes, that's why I was thinking adding "INCLUDE" or something like that will help. Anyway, it's fine, as log as a flag is there, it should be fine.

ashokponkumar (approved):

PR #451, opened 2021-03-09
fix: javamaven dockerfile template change

alealavi:
app folder was not being created, image was changed for openjdk 1.8 and works, only needed is change the .app_name to be the same as the current project name being translated


HarikrishnanBalagopal (dismissed):
Need to run `make generate` for this change to have any effect.

The original issue may have been because of some other reason.

alealavi:
you were right it was probably a permissions problem since now it works without having to make the app folder, however I still get a jboss error if I leave the image, so Im leaving that change but reverted the "mkdir app" one

I also signed off the changes this time so the checks all pass and generated a new constants file

The jboss error should be below if you want to see it, but with the image change now it works

image

HarikrishnanBalagopal:
Sure, the base image change seems fine.

Thanks for making the changes!

HarikrishnanBalagopal (approved):
LGTM

PR #470, opened 2021-04-06
feat: always copy sources to the source folder

HarikrishnanBalagopal:
No real benefit in adding an extra step to copy
sources. UI already copies sources anyway.


ashokponkumar (changes requested):
We should list the files that have been added before copying over. Otherwise it will be hard for the users to identify the new files.

ashokponkumar (dismissed):

ashokponkumar (approved):

PR #480, opened 2021-04-27
fix: helm chart folder name should be same as the chart name

HarikrishnanBalagopal:

### DIFF
--- a/internal/transformer/k8stransformer.go
+++ b/internal/transformer/k8stransformer.go
@@ -199,7 +199,7 @@ func (kt *K8sTransformer) WriteObjects(outputPath string, transformPaths []strin
 	kt.writeDeployScript(kt.Name, outputPath)
 
 	// deploy/operator/
-	if err := kt.createOperator(kt.Name, filepath.Join(deployPath, "operator"), filepath.Join(deployPath, "helm")); err != nil {
+	if err := kt.createOperator(kt.Name, filepath.Join(deployPath, "operator"), filepath.Join(deployPath, common.HelmDir)); err != nil {
 		log.Errorf("Failed to generate the operator. Error: %q", err)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/transformer/k8stransformer.go
```diff
@@ -218,9 +214,9 @@ func (kt *K8sTransformer) WriteObjects(outputPath string, transformPaths []strin
 	return nil
 }
 
-func (kt *K8sTransformer) generateHelmMetadata(outputPath string, values outputtypes.HelmValues) error {
+func (kt *K8sTransformer) generateHelmArtifacts(outputPath string, values outputtypes.HelmValues, transformPaths []string) error {
 	deployPath := filepath.Join(outputPath, common.DeployDir)
-	helmPath := filepath.Join(deployPath, "helm")
+	helmPath := filepath.Join(deployPath, "helm", kt.Name)
```
ashokponkumar, line 219: Lets call it "helm-charts". Probably we can add a helm chart for the ci/cd artifacts too.

ashokponkumar (changes requested):
Also, What is the folder structure for kustomize artifacts? is it good to have the folder name there too?

Should we put ci/cd inside each of these yamls, kustomize and helm too, so that it allows for install of ci/cd using them too?

HarikrishnanBalagopal:
> 
> 
> Also, What is the folder structure for kustomize artifacts? is it good to have the folder name there too?
> 
> Should we put ci/cd inside each of these yamls, kustomize and helm too, so that it allows for install of ci/cd using them too?

https://github.com/konveyor/move2kube/blob/main/internal/transformer/k8stransformer.go#L160-L176

internal/transformer/k8stransformer.go
HarikrishnanBalagopal: We already have CI/CD don't we?

HarikrishnanBalagopal: `Lets call it "helm-charts"` done

ashokponkumar (approved):

PR #482, opened 2021-04-27
fix: Update helm install command with new path

ashokponkumar:


HarikrishnanBalagopal (approved):
LGTM

PR #485, opened 2021-04-29
feat: generate openshift templates

HarikrishnanBalagopal:


ashokponkumar (approved):

internal/transformer/k8stransformer.go
```diff
@@ -395,3 +403,39 @@ func (kt *K8sTransformer) writeReadMe(project string, areNewImages bool, outpath
 		log.Errorf("Unable to write readme : %s", err)
 	}
 }
+
+func (kt *K8sTransformer) generateOpenshiftTemplates(ocTemplatesPath, outputPath string, objs []runtime.Object) ([]string, error) {
+	// deploy/openshift-templates/
+	templ := new(templatev1.Template)
+	templ.TypeMeta.APIVersion = "template.openshift.io/v1"
```
ashokponkumar, line 410: Do we want to use the template { Val:...
Val2:..
} Method to initialise, so that we need not have multiple assignments.

ashokponkumar (changes requested):

internal/transformer/k8stransformer.go
HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #491, opened 2021-05-06
fix: Resolve the path bug while exposing a service with multiple ports

Akash-Nayak:


ashokponkumar:
Fixed using per port questions.

PR #492, opened 2021-05-07
feat: Added support for multiple ports in dockerfilecontainerizer

seshapad:
Signed-off-by: Padmanabha Venkatagiri Seshadri <seshapad@in.ibm.com>


internal/assets/constants.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
...
...AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIyEvdXNyL2Jpbi9lbnYgYmFzaAojICAgQ29weXJpZ2h0IElCTSBDb3Jwb3JhdGlvbiAyMDIwCiMKIyAgIExpY2Vuc2VkIHVuZGVyIHRoZSBBcGFjaGUgTGljZW5zZSwgVmVyc2lvbiAyLjAgKHRoZSAiTGljZW5zZSIpOwojICAgeW91IG1heSBub3QgdXNlIHRoaXMgZmlsZSBleGNlcHQgaW4gY29tcGxpYW5jZSB3aXRoIHRoZSBMaWNlbnNlLgojICAgWW91IG1heSBvYnRhaW4gYSBjb3B5IG9mIHRoZSBMaWNlbnNlIGF0CiMKIyAgICAgICAgaHR0cDovL3d3dy5hcGFjaGUub3JnL2xpY2Vuc2VzL0xJQ0VOU0UtMi4wCiMKIyAgIFVubGVzcyByZXF1aXJlZCBieSBhcHBsaWNhYmxlIGxhdyBvciBhZ3JlZWQgdG8gaW4gd3JpdGluZywgc29mdHdhcmUKIyAgIGRpc3RyaWJ1dGVkIHVuZGVyIHRoZSBMaWNlbnNlIGlzIGRpc3RyaWJ1dGVkIG9uIGFuICJBUyBJUyIgQkFTSVMsCiMgICBXSVRIT1VUIFdBUlJBTlRJRVMgT1IgQ09ORElUSU9OUyBPRiBBTlkgS0lORCwgZWl0aGVyIGV4cHJlc3Mgb3IgaW1wbGllZC4KIyAgIFNlZSB0aGUgTGljZW5zZSBmb3IgdGhlIHNwZWNpZmljIGxhbmd1YWdlIGdvdmVybmluZyBwZXJtaXNzaW9ucyBhbmQKIyAgIGxpbWl0YXRpb25zIHVuZGVyIHRoZSBMaWNlbnNlLgoKIyBUYWtlcyBhcyBpbnB1dCB0aGUgc291cmNlIGZvbGRlciBhbmQgcmV0dXJucyBlcnJvciBpZiBpdCBpcyBub3QgZml0CkJBU0VfRElSPSQxCklNQUdFPSJyZWdpc3RyeS5hY2Nlc3MucmVkaGF0LmNvbS9yaHNjbC9ydWJ5LTI1LXJoZWw3OmxhdGVzdCIKCmlmIFsgISAtZiAiJEJBU0VfRElSIi9HZW1maWxlIF07IHRoZW4KICAgIGV4aXQgMQpmaQoKZWNobyAneyJidWlsZGVyIjogIickSU1BR0UnIiwgInBvcnQiOiA4MDgwfScK`
```
ashokponkumar, line 21: Is there any changes in the constants? I think we don't need this change.

internal/containerizer/dockerfilecontainerizer.go
```diff
@@ -182,6 +183,19 @@ func (d *DockerfileContainerizer) GetContainer(plan plantypes.Plan, service plan
 			container.AddExposedPort(portToExpose)
 		}
 
+		// is "ports" present ?
+		if val, ok := segmentRecord["ports"]; ok {
+			switch reflect.TypeOf(val).Kind() {
```
ashokponkumar, line 188: Since it is only one type check, would it be better to use `if` instead of switch case?

Would it be better to use a check like val.([]int) to get this value, so that it does both the type check and also get the value in the right format in one go?

ashokponkumar (changes requested):

ashokponkumar:
Also please add `feat:` to the title of the PR, since you are adding a new feature.

HarikrishnanBalagopal:
Seems to be invalid for main branch. Closing.
If it is relevant for v0.2.0 create a new PR targetting `release-0.2`

PR #529, opened 2021-06-29
feat: add the parameterize command

HarikrishnanBalagopal:
Parameterizer allows parameterizing fields in the k8s resources.
Also ignore the ruby dependency directories.


newparameterizer/common/utils.go
```diff
@@ -0,0 +1,448 @@
...
+func WriteResourceAppendToFile(k8sResource types.K8sResourceT, outputPath string) error {
+	logrus.Trace("start WriteResourceAppendToFile")
+	defer logrus.Trace("end WriteResourceAppendToFile")
+	yamlBytes, err := yaml.Marshal(k8sResource)
+	if err != nil {
+		logrus.Error("Error while Encoding object")
+		return err
+	}
+	// If the file doesn't exist, create it, or append to the file
+	f, err := os.OpenFile(outputPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, common.DefaultFilePermission)
```
ashokponkumar, line 417: Do we want to maintain the same permissions and files as the original directory. For example, if a multiple artifacts are present in a single file, we could possibly retain the same structure for helm and possibly for kustomize too?

ashokponkumar (changes requested):
Nice work and Nice implementation. Have added a few comments, the major one is rename the package to parameterizer instead of newparameterizer. Other than that have made some small suggestions. Other fixes/changes we can continue to improve over time, as we start using the feature.

PR #547, opened 2021-07-06
fix: Make only lower case container image names

ashokponkumar:


ashokponkumar:
This will be available in v0.3+ builds

PR #556, opened 2021-07-13
feat: parameter extraction and image selection

pabloloyola:


assets/inbuilt/transformers/springboot/java-springboot/mappings/java2images_tags.json
```diff
@@ -0,0 +1 @@
...ss/wildfly:21.0.0.Final", "created": "2020-10-21T09:25:31.897267464Z", "architecture": "amd64"}, {"name": "jboss/wildfly:22.0.1.Final", "created": "2021-02-12T08:21:48.823067126Z", "architecture": "amd64"}, {"name": "jboss/wildfly:15.0.0.Final", "created": "2019-02-13T13:53:43.861482142Z", "architecture": "amd64"}, {"name": "jboss/wildfly:23.0.1.Final", "created": "2021-04-15T07:42:10.113693208Z", "architecture": "amd64"}, {"name": "jboss/wildfly:21.0.1.Final", "created": "2020-12-02T12:57:04.468022499Z", "architecture": "amd64"}, {"name": "jboss/wildfly:19.0.0.Final", "created": "2020-03-27T10:08:38.660818401Z", "architecture": "amd64"}, {"name": "jboss/wildfly:16.0.0.Final", "created": "2020-09-07T09:21:27.707384617Z", "architecture": "amd64"}, {"name": "jboss/wildfly:19.1.0.Final", "created": "2020-05-31T19:09:37.883038006Z", "architecture": "amd64"}, {"name": "jboss/wildfly:17.0.0.Final", "created": "2019-06-11T07:37:44.87732309Z", "architecture": "amd64"}, {"name": "jboss/wildfly:20.0.1.Final", "created": "2020-09-07T09:25:05.065383866Z", "architecture": "amd64"}, {"name": "jboss/wildfly:18.0.0.Final", "created": "2019-10-07T08:07:41.180078006Z", "architecture": "amd64"}]}]}]}
```
ashokponkumar, line 1: The structure of this json seems to be custom made for this very specific implementation. If we have to say add more filters other than java version or app server, we need another json. Would it be better if we flatten it out, and then have a list of images and properties for each image, which will include app server name, java version and more. This way, we can add any more meta data to the images without having to worry about the hierarchy.

ashokponkumar (changes requested):
Nice work. A few change suggestions.

ashokponkumar (approved):

PR #576, opened 2021-08-07
feat: transformers for windows dotnet apps

seshapad:


internal/transformer/transformer.go
```diff
@@ -80,6 +80,9 @@ func init() {
 		new(generators.ReadMeGenerator),
 
 		new(dockerfilegenerators.NodejsDockerfileGenerator),
+		new(dockerfilegenerators.WinConsoleAppDockerfileGenerator),
```
ashokponkumar, line 83: Would be good to put all dotnet ones in a new package below dockerfilegenerators to keep them together.

assets/inbuilt/transformers/dockerfile/winconsole/winconsole.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: move2kube.konveyor.io/v1alpha1
+kind: Transformer
+metadata:
+  name: WinConsoleApp-Dockerfile
+spec:
+  mode: "Container"
+  class: "WinConsoleAppDockerfileGenerator"
+  generates:
```
ashokponkumar, line 8: The generates and generatesbases fields are no longer part of the yaml. We can remove them

assets/inbuilt/transformers/dockerfile/winslweb/winslweb.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: move2kube.konveyor.io/v1alpha1
+kind: Transformer
+metadata:
+  name: WinSLWebApp-Dockerfile
```
ashokponkumar, line 4: Sl is not very obvious. Use silverlight

assets/inbuilt/transformers/dockerfile/winslweb/winslweb.yaml
```diff
@@ -0,0 +1,11 @@
...
+spec:
+  mode: "Container"
+  class: "WinSLWebAppDockerfileGenerator"
+  generates:
```
ashokponkumar, line 8: Remove generates and generatebases

assets/inbuilt/transformers/dockerfile/winweb/winweb.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: move2kube.konveyor.io/v1alpha1
+kind: Transformer
+metadata:
+  name: WinWebApp-Dockerfile
+spec:
+  mode: "Container"
+  class: "WinWebAppDockerfileGenerator"
+  generates:
```
ashokponkumar, line 8: Remove generates and generatedbases

internal/transformer/classes/generators/dockerfilegenerators/utils.go
```diff
@@ -17,15 +17,108 @@
...
 
 	"github.com/konveyor/move2kube/environment"
 	"github.com/konveyor/move2kube/internal/common"
+	"github.com/konveyor/move2kube/types/source/dotnet"
 	transformertypes "github.com/konveyor/move2kube/types/transformer"
 	"github.com/konveyor/move2kube/types/transformer/artifacts"
 	"github.com/sirupsen/logrus"
 )
 
+// isSilverlight tests whether the app is a sliverlight web application
```
ashokponkumar, line 35: Let's move this to a separate subpackage along with other dotnet Transformers.

internal/transformer/classes/generators/dockerfilegenerators/utils.go
```diff
@@ -17,15 +17,108 @@
...
+
+		for _, r := range ig.References {
+			if dotnet.WebLib.MatchString(r.Include) {
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
+}
```
ashokponkumar, line 75: Better to return error too from this function.

internal/transformer/classes/generators/dockerfilegenerators/utils.go
```diff
@@ -17,15 +17,108 @@
...
+
+// parseSolutionFile parses the solution file for the project file path
+func parseSolutionFile(inputPath string) []string {
+	solFile, err := os.Open(inputPath)
+	if err != nil {
+		logrus.Errorf("Could not open the solution file: %s", err)
+		return nil
+	}
+	defer solFile.Close()
+
```
ashokponkumar, line 85: Would be better to store the compiled refer in a variable rather than compiling it everytime

internal/transformer/classes/generators/dockerfilegenerators/utils.go
```diff
@@ -17,15 +17,108 @@
...
+	solFileScanner := bufio.NewScanner(solFile)
+
+	r := regexp.MustCompile(dotnet.ProjBlock)
+
+	csr := regexp.MustCompile(dotnet.CsProj)
```
ashokponkumar, line 90: If the solution file is an XML, why not use a XML parser?

internal/transformer/classes/generators/dockerfilegenerators/utils.go
```diff
@@ -17,15 +17,108 @@
...
+				}
+			}
+		}
+	}
+
+	if err := solFileScanner.Err(); err != nil {
+		logrus.Errorf("Could not parse the solution file: %s", err)
+		return nil
+	}
+
```
ashokponkumar, line 112: Won't this depend on the is the progem is in? If so, we should avoid hardcoding the slash type

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// BaseDirectoryDetect runs detect in base directory
+func (t *WinConsoleAppDockerfileGenerator) BaseDirectoryDetect(dir string) (namedServices map[string]transformertypes.ServicePlan, unnamedServices []transformertypes.TransformerPlan, err error) {
+	return nil, nil, nil
+}
+
+// parseAppConfig parses the app.config file for ports
+func (t *WinConsoleAppDockerfileGenerator) parseAppConfig(baseDir string) []string {
+	appConfigFile, err := os.Open(filepath.Join(baseDir, "App.config"))
```
ashokponkumar, line 60: Won't it be better to return error in this case

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		logrus.Errorf("Could not parse the App.config file: %s", err)
+		return nil
+	}
+
+	ports := make([]string, 0)
+	for _, addKey := range appCfg.AppCfgSettings.AddList {
+		if parsedUrl, err := url.ParseRequestURI(addKey.Value); err == nil && parsedUrl.Scheme != "" && parsedUrl.Host != "" {
+			_, port, _ := net.SplitHostPort(parsedUrl.Host)
+			if err == nil {
```
ashokponkumar, line 79: Which err is this checking? We should at least log a message in the err case

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+				ports = append(ports, port)
+			}
+		}
+	}
+
+	if len(appCfg.Model.Services.ServiceList) > 0 {
```
ashokponkumar, line 85: Would be better to check ==0 and continue to reduce indentation

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+		for _, svc := range appCfg.Model.Services.ServiceList {
+			for _, addKey := range svc.Host.BaseAddresses.AddList {
+				if parsedUrl, err := url.ParseRequestURI(addKey.BaseAddress); err == nil && parsedUrl.Scheme != "" && parsedUrl.Host != "" {
+					_, port, _ := net.SplitHostPort(parsedUrl.Host)
```
ashokponkumar, line 89: Isn't this supposed to return an error too

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+	dirEntries, err := os.ReadDir(dir)
+	if err != nil {
+		logrus.Errorf("Error while trying to read directory: %s", err)
+		return nil, nil, err
+	}
+	appName := ""
+	ports := make([]string, 0)
+	for _, de := range dirEntries {
+		ext := filepath.Ext(de.Name())
+		if ext != dotnet.CsSln {
```
ashokponkumar, line 112: If ext value is used only here, no need of a variable

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+			continue
+		}
+		csProjPaths := parseSolutionFile(filepath.Join(dir, de.Name()))
+
+		if csProjPaths == nil || len(csProjPaths) == 0 {
```
ashokponkumar, line 117: Pen checks for nil too, so no need of a separate nil check

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
```diff
@@ -0,0 +1,185 @@
...
+			if err != nil || isWebProj {
+				continue
+			}
+
+			portList := t.parseAppConfig(filepath.Join(dir, filepath.Dir(csPath)))
+			if portList != nil {
+				ports = append(ports, portList...)
+			}
+
+			appName = strings.TrimSuffix(filepath.Base(de.Name()), filepath.Ext(de.Name()))
```
ashokponkumar, line 154: If there is more than one solution file, the behaviour is undeterministic now. We should think about how we should act in that case

types/source/dotnet/csproj.go
```diff
@@ -0,0 +1,133 @@
...
+
+var (
+	// FourXPattern is the framework version pattern
+	FourXPattern = regexp.MustCompile("v4*")
+
+	// WebLib is the key library used in web applications
+	WebLib = regexp.MustCompile("^System.Web*")
+
+	// WebSLLib is the key library used in web sliverlight applications
+	WebSLLib = regexp.MustCompile("Silverlight.js")
```
ashokponkumar, line 33: If we are using it as regex, the dot need to be escaped

types/source/dotnet/csproj.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"encoding/xml"
+	"io"
+	"regexp"
+)
+
+var (
+	// FourXPattern is the framework version pattern
+	FourXPattern = regexp.MustCompile("v4*")
```
ashokponkumar, line 27: This is not a valid regex did we want .* in the end?

types/source/dotnet/csproj.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// WebLib is the key library used in web applications
+	WebLib = regexp.MustCompile("^System.Web*")
```
ashokponkumar, line 30: This also seems to be not the intended regex. It is looking for 0 or more b's in the end

ashokponkumar (changes requested):
Nice work @seshapad . Looking forward to getting windows support natively added.

Have done one pass and have added some comments.

internal/transformer/classes/generators/dockerfilegenerators/utils.go
seshapad: @ashokponkumar  solution files are text files, not xml...they have their own unique format .

ashokponkumar: Okay sure. @pabloloyola do we know if there is a antlr4 parser for visual studio solution file

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/Akash-Nayak/GopacheConfig v0.0.0-20210730101443-d5bfa3109be4
 	github.com/AlecAivazis/survey/v2 v2.2.12
 	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/apex/log v1.3.0 // indirect
```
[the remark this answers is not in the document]
seshapad: done

internal/transformer/classes/generators/dockerfilegenerators/winslwebappdockerfilegenerator.go
```diff
@@ -0,0 +1,139 @@
...
+	dirEntries, err := os.ReadDir(dir)
+	if err != nil {
+		logrus.Errorf("Error while trying to read directory: %s", err)
+		return nil, nil, err
+	}
+	appName := ""
+	ports := make([]string, 0)
+	for _, de := range dirEntries {
+		ext := filepath.Ext(de.Name())
+		if ext != dotnet.CsSln {
```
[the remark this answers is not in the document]
seshapad: done

internal/transformer/transformer.go
seshapad: done

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
seshapad: @ashokponkumar More than one solution file for project is highly unlikely.

internal/transformer/classes/generators/dockerfilegenerators/utils.go
seshapad: @ashokponkumar could you clarify this comment?

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
seshapad: I have logged the error.

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
ashokponkumar: Sure, In that case we should break after processing a single solution file.

seshapad: done

internal/transformer/classes/generators/dockerfilegenerators/utils.go
seshapad: @ashokponkumar I have covered all the error logging within the function itself. Is that an issue? whats the best practice.

internal/transformer/classes/generators/dockerfilegenerators/utils.go
seshapad: @ashokponkumar the relative path that comes from the csproj file has windows-slashes...but if i am running move2kube on mac, but trying to read the csproj, then I get the base path in mac-slash format, and the relative path in win-slash format...hence converting them both to a single format to ease-of-use. Is this approach ok?

seshapad:
@ashokponkumar I have signed off all commits and DCO shows not errors as shown below...but the build is still failing on DCO as seen above. What could be the issue?
image

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
ashokponkumar: When the function is not able to perform the primary job it is supposed to do, then it is better to return error, and the caller has to decide how to proceed further.  Since in this function it is about parsing a config file, and the function is not even able to open the file, it is better to return error and log/handle it in the caller code.

ashokponkumar:
The build is not failing on DCO

internal/transformer/classes/generators/dockerfilegenerators/winconsoleappdockerfilegenerator.go
seshapad: Fixed

internal/transformer/classes/generators/dockerfilegenerators/utils.go
seshapad: fixed

seshapad:
Closing this PR to open a new one.

PR #585, opened 2021-09-07
fix:  Naming issues for container block and network policy labels

seshapad:

### DIFF
--- a/internal/transformer/classes/analysers/compose/v3.go
+++ b/internal/transformer/classes/analysers/compose/v3.go
@@ -158,7 +158,7 @@ func (c *V3Loader) convertToIR(filedir string, composeObject types.Config, servi
 		serviceContainer.Command = composeServiceConfig.Entrypoint
 		serviceContainer.Args = composeServiceConfig.Command
 		serviceContainer.Stdin = composeServiceConfig.StdinOpen
-		serviceContainer.Name = strings.ToLower(composeServiceConfig.ContainerName)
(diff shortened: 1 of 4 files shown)


internal/transformer/classes/analysers/compose/v3.go
```diff
@@ -158,7 +158,7 @@ func (c *V3Loader) convertToIR(filedir string, composeObject types.Config, servi
 		serviceContainer.Command = composeServiceConfig.Entrypoint
 		serviceContainer.Args = composeServiceConfig.Command
 		serviceContainer.Stdin = composeServiceConfig.StdinOpen
-		serviceContainer.Name = strings.ToLower(composeServiceConfig.ContainerName)
+		serviceContainer.Name = common.NormalizeForServiceName(composeServiceConfig.ContainerName)
```
ashokponkumar, line 161: Do we want to create a new function for normalizing container names? It might have different requirements? Also, does it need to do ToLower?

ashokponkumar (changes requested):

internal/transformer/classes/analysers/compose/v3.go
seshapad: I am using an existing function common.NormalizeForServiceName which does all this. Container names are lowercase by default. I remember getting error from docker cli when camel-case was used.

ashokponkumar (approved):

PR #592, opened 2021-09-23
feat: add a REST server to report the progress of the planning phase

HarikrishnanBalagopal:


cmd/move2kube/plan.go
```diff
@@ -125,7 +140,10 @@ func getPlanCommand() *cobra.Command {
 	planCmd.Flags().StringSliceVar(&flags.preSets, preSetFlag, []string{}, "Specify preset config to use")
 	planCmd.Flags().StringArrayVar(&flags.setconfigs, setConfigFlag, []string{}, "Specify config key-value pairs")
 
+	planCmd.Flags().IntVar(&flags.progressServerPort, "plan_progress_port", 0, "Port for the plan progress server. If not provided, the server won't be started.")
+
 	must(planCmd.MarkFlagRequired(sourceFlag))
+	must(planCmd.Flags().MarkHidden("plan_progress_port"))
```
ashokponkumar, line 146: Can we move `plan_progress_port` as a flags file similar to our other flag names? Also, till we refactor other flag names to keep it consistent, lets use `planprogressport`. Its not ideal, but would be good to keep standard. Or we can refactor all flag names.

internal/transformer/transformer.go
```diff
@@ -329,6 +329,7 @@ func walkForServices(inputPath string, ts map[string]Transformer, bservices map[
 	knownProjectPaths := []string{}
 
 	err = filepath.Walk(inputPath, func(path string, info os.FileInfo, err error) error {
+		common.PlanProgressNumFiles++
```
ashokponkumar, line 332: We might want to use atomic for this. Otherwise we might end up with panic. https://gobyexample.com/atomic-counters

ashokponkumar (changes requested):
Looks good overall. Have added a few suggestions.

ashokponkumar (approved):

PR #596, opened 2021-09-30
feat: Use docker or podman to build and run

rmarting:
This PR improves the Makefile to check which container tool is installed in the enviroment (only docker or podman).

The user will not have to set up anything before to build or to run the image locally, only a container tool is needed to be installed.

HTH


rmarting:
@ashokponkumar Here the PR as you request me related with https://github.com/konveyor/move2kube-ui/pull/98

Ready for your review.

ashokponkumar (approved):

PR #598, opened 2021-10-05
feat: Added write capability to starlark script

seshapad:


internal/transformer/classes/external/starlark.go
```diff
@@ -291,7 +339,8 @@ func (t *Starlark) addAppModules() {
 	t.StarGlobals[types.AppNameShort] = &starlarkstruct.Module{
 		Name: types.AppNameShort,
 		Members: starlark.StringDict{
-			qaFnName: t.getStarlarkQuery(),
+			qaFnName:    t.getStarlarkQuery(),
+			writeFnName: t.getStarlarkFileWrite(),
```
ashokponkumar, line 343: This function should come under fs module. Can we add it inside addFSModules function?

internal/transformer/classes/external/starlark.go
```diff
@@ -75,6 +76,12 @@ type StarYamlConfig struct {
 	StarFile string `yaml:"starFile"`
 }
 
+// WriteFnParams defines yaml config for Starlark transformers
```
ashokponkumar, line 79: Do we need a yaml config for this function?

internal/transformer/classes/external/starlark.go
```diff
@@ -75,6 +76,12 @@ type StarYamlConfig struct {
...
+type WriteFnParams struct {
+	FilePath string `yaml:"FilePath" json:"FilePath"`
+	Data     []byte `yaml:"Data" json:"Data"`
```
ashokponkumar, line 82: Can't this be taken as two parameters in the function?

internal/transformer/classes/external/starlark.go
```diff
@@ -252,6 +259,47 @@ func (t *Starlark) getStarlarkQuery() *starlark.Builtin {
 	})
 }
 
+func (t *Starlark) getStarlarkFileWrite() *starlark.Builtin {
```
ashokponkumar, line 262: Are we not using the os starlark library you had mentioned?

ashokponkumar (changes requested):
Thanks @seshapad. Few questions...

PR #640, opened 2021-11-19
feat: Add golang-based transformer for Python and Django

Akash-Nayak:
Added transformer for Python and Django- https://github.com/konveyor/move2kube/issues/628, https://github.com/konveyor/move2kube/issues/629


transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+			continue
+		}
+		sImageName := artifacts.ImageName{}
+		err = a.GetConfig(artifacts.ImageNameConfigType, &sImageName)
+		if err != nil {
+			logrus.Debugf("unable to load config for Transformer into %T : %s", sImageName, err)
+		}
+		var pythonConfig PythonTemplateConfig
+		pythonConfig.AppName = a.Name
+		if a.Name == djangoService {
```
ashokponkumar, line 155: This should be passed as part of a config in artifact and should not be inferred from service name.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+		mainPythonFileRelPath, err := filepath.Rel(a.Paths[artifacts.ProjectPathPathType][0], a.Paths[MainPythonFilePathType][0])
+		if err != nil {
+			logrus.Errorf("Error in getting the relative path %s", err)
+		} else {
+			pythonConfig.MainScriptRelPath = mainPythonFileRelPath
+		}
+		var detectedPorts []int32
+		detectedPorts = append(detectedPorts, 8080) //TODO: Write parser to parse and identify port
+		pythonConfig.Port = commonqa.GetPortForService(detectedPorts, a.Name)
+		requirementsTxtFiles, err := common.GetFilesByName(a.Paths[artifacts.ProjectPathPathType][0], []string{requirementsTxtFile}, nil)
```
ashokponkumar, line 167: We can detect all these as part of the detect function as pass it as part of the artifact.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+			Type:     transformertypes.SourcePathMappingType,
+			DestPath: common.DefaultSourceDir,
+		}, transformertypes.PathMapping{
+			Type:           transformertypes.TemplatePathMappingType,
+			SrcPath:        filepath.Join(t.Env.Context, t.Config.Spec.TemplatesDir),
+			DestPath:       filepath.Join(common.DefaultSourceDir, relSrcPath),
+			TemplateConfig: pythonConfig,
+		})
+		paths := a.Paths
+		paths[artifacts.DockerfilePathType] = []string{filepath.Join(common.DefaultSourceDir, relSrcPath, "Dockerfile")}
```
ashokponkumar, line 192: Do we want to use the const for Dockerfilename

ashokponkumar (changes requested):
Thanks @Akash-Nayak . Few suggestions.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+	for _, pythonFilePath := range pythonFilesPath {
+		pythonFile, err := os.Open(pythonFilePath)
+		if err != nil {
+			logrus.Debugf("Failed to open the file %s", pythonFilePath)
+		}
+		scanner := bufio.NewScanner(pythonFile)
+		scanner.Split(bufio.ScanLines)
+		for scanner.Scan() {
+			if pythonMainRegex.MatchString(scanner.Text()) {
+				return pythonFilePath, nil
```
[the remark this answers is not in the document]
Akash-Nayak: Yes sure. Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, err
+	}
+	// Search requirements.txt
+	if mainPythonFilePath != "" {
+		serviceName := strings.TrimSuffix(filepath.Base(mainPythonFilePath), pythonExt)
+		if serviceName == pythonMain {
+			serviceName = pythonService
+		} else if serviceName == pythonManage {
+			serviceName = djangoService
+		}
```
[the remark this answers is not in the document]
Akash-Nayak: No, we can leave it as unnamed service and let move2kube decide the name of the service.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// Search requirements.txt
+	if mainPythonFilePath != "" {
+		serviceName := strings.TrimSuffix(filepath.Base(mainPythonFilePath), pythonExt)
+		if serviceName == pythonMain {
+			serviceName = pythonService
+		} else if serviceName == pythonManage {
+			serviceName = djangoService
```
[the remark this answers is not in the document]
Akash-Nayak: In general, what I noticed was that all the django projects have this `manage.py` file, so I was using that for deciding whether it's a django project or not. But instead of doing that, now we are using the `requirements.txt` file and checking if `django` is present in it or not.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,213 @@
...
+				return pythonFilePath, nil
+			}
+		}
+	}
+	return "", fmt.Errorf("could not find the main function in python files")
+}
+
+// DirectoryDetect runs detect in each sub directory
+func (t *PythonDockerfileGenerator) DirectoryDetect(dir string) (services map[string][]transformertypes.Artifact, err error) {
+	pythonFiles, err := common.GetFilesByExt(dir, []string{pythonExt})
```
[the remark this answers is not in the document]
Akash-Nayak: Currently, it's working, i.e., detecting other projects as well, if I keep `.m2kignore`  file in the input directory.

I think that currently GetFilesByName also searches for the filenames in all sub-directories. For example:

```
common.GetFilesByName(dir, []string{"requirements.txt", "urls.py", "manage.py"}, nil)

Output:
/var/folders/sw/_kjrz98j3ws1qj7zz6chzhrr0000gn/T/move2kube1506205050/environment-Python-Dockerfile-381637603/workspace3434827603/django/manage.py
/var/folders/sw/_kjrz98j3ws1qj7zz6chzhrr0000gn/T/move2kube1506205050/environment-Python-Dockerfile-381637603/workspace3434827603/django/requirements.txt
/var/folders/sw/_kjrz98j3ws1qj7zz6chzhrr0000gn/T/move2kube1506205050/environment-Python-Dockerfile-381637603/workspace3434827603/django/simplewebapp/urls.py
/var/folders/sw/_kjrz98j3ws1qj7zz6chzhrr0000gn/T/move2kube1506205050/environment-Python-Dockerfile-381637603/workspace3434827603/django/webroot/urls.py
```
Here, it detected the `urls.py` present inside the sub-folders.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Yes sure :). Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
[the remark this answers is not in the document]
ashokponkumar: We should require m2kignore only as an exception. In this case itight be good to check for py files only in the current directory.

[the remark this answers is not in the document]
Akash-Nayak: Okay sure.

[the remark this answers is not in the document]
Akash-Nayak: I have added GetFilesByExtInCurrDir function which will only look for files in the current directory.

common/utils.go
```diff
@@ -92,6 +92,39 @@ func GetFilesByExt(inputPath string, exts []string) ([]string, error) {
...
 }
 
+// GetFilesByExtInCurrDir returns the files present in current directory which have one of the specified extensions
+func GetFilesByExtInCurrDir(inputPath string, exts []string) ([]string, error) {
+	var files []string
+	if info, err := os.Stat(inputPath); os.IsNotExist(err) {
+		logrus.Warnf("Error in walking through files due to : %q", err)
+		return nil, err
+	} else if !info.IsDir() {
+		logrus.Warnf("The path %q is not a directory.", inputPath)
```
ashokponkumar, line 102: Should we not just check the extension of the file here and return? If not, won't readdir command fail here?

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+	RequirementsTxtConfigType transformertypes.ConfigType = "RequirementsTxtConfigType"
+	// DjangoProjectConfigType is set to true for Django projects
+	DjangoProjectConfigType transformertypes.ConfigType = "DjangoProjectConfigType"
+	// MainPythonFilesPathType points to the .py file path which contains main function
+	MainPythonFilesPathType transformertypes.PathType = "MainPythonFilesPathType"
+)
+
+var (
+	pythonMainRegex       = regexp.MustCompile(`__main__`)
+	djangoDependencyRegex = regexp.MustCompile(`django`)
```
ashokponkumar, line 54: Won't a contains command work in both these cases? do we need regex?

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/konveyor/move2kube/types/transformer/artifacts"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	pythonExt           = ".py"
+	pythonMain          = "main"
+	pythonManage        = "manage"
+	pythonService       = "python"
+	djangoService       = "django"
```
ashokponkumar, line 40: Are we using these consts now?

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+		logrus.Debugf("Failed to open the file %s", reqTxtFilePath)
+	}
+	scanner := bufio.NewScanner(reqTxtFile)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		if djangoDependencyRegex.MatchString(scanner.Text()) {
+			return true
+		}
+	}
+	return false
```
ashokponkumar, line 122: won't a ioutil.Readfile and strings.contains be sufficient?

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+	//RequirementsTxtConfigType points to the requirements.txt file if it's present
+	RequirementsTxtConfigType transformertypes.ConfigType = "RequirementsTxtConfigType"
+	// DjangoProjectConfigType is set to true for Django projects
+	DjangoProjectConfigType transformertypes.ConfigType = "DjangoProjectConfigType"
+	// MainPythonFilesPathType points to the .py file path which contains main function
+	MainPythonFilesPathType transformertypes.PathType = "MainPythonFilesPathType"
+)
+
+var (
+	pythonMainRegex       = regexp.MustCompile(`__main__`)
```
ashokponkumar, line 53: Shouldn't we be looking for the entire variable assignment `__name__ == '__main__'`? If not, even a if __name__ == '__main__' line would succeed. Isn't it? Like in https://docs.python.org/3/library/__main__.html

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+	mainPythonFilesPath, err := findMainScripts(pythonFiles)
+	if err != nil {
+		return nil, err
+	}
+	if len(mainPythonFilesPath) != 0 {
+		var serviceName, requirementsTxtPath string
+		var isDjangoProject bool
+		var detectedPorts []int32
+		detectedPorts = append(detectedPorts, 8080) //TODO: Write parser to parse and identify port
+		requirementsTxtFiles, err := common.GetFilesByName(dir, []string{requirementsTxtFile}, nil)
```
ashokponkumar, line 156: Shouldn't we be using GetFilesInCurrentDirectory? Generally requirements.txt is placed only in the base directory of the project. isn't it?

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+		if err != nil {
+			logrus.Debugf("Cannot get the requirements.txt file: %s", err)
+		}
+		if len(requirementsTxtFiles) == 1 {
```
ashokponkumar, line 160: Seems like using requirements directory is another general pattern : https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+		}
+		services = map[string][]transformertypes.Artifact{
+			serviceName: {{
+				Paths: map[string][]string{
+					artifacts.ProjectPathPathType: {dir},
+					MainPythonFilesPathType:       mainPythonFilesPath,
+				}, Configs: map[string]interface{}{
+					DjangoProjectConfigType:   isDjangoProject,
+					RequirementsTxtConfigType: requirementsTxtPath,
```
ashokponkumar, line 176: Paths should go in Paths and not in Configs.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// Transform transforms the artifacts
+func (t *PythonDockerfileGenerator) Transform(newArtifacts []transformertypes.Artifact, oldArtifacts []transformertypes.Artifact) ([]transformertypes.PathMapping, []transformertypes.Artifact, error) {
+	pathMappings := []transformertypes.PathMapping{}
+	artifactsCreated := []transformertypes.Artifact{}
+	for _, a := range newArtifacts {
+		relSrcPath, err := filepath.Rel(t.Env.GetEnvironmentSource(), a.Paths[artifacts.ProjectPathPathType][0])
+		if err != nil {
+			logrus.Errorf("Unable to convert source path %s to be relative : %s", a.Paths[artifacts.ProjectPathPathType][0], err)
```
ashokponkumar, line 193: Shouldn't we continue to the next artifact here?

ashokponkumar (changes requested):
Nice work Akash! Few more suggestions.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: No, I have removed them.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Yes, contains command will work in both the cases. So, we don't need regex. 
Done.

common/utils.go
Akash-Nayak: Yes, we should check the extension of the file here. Thanks for detecting the corner case. 
Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Sure, I have changed it to the GetFilesInCurrentDirectory to look for the file only in the base directory.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Done.

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: As discussed yesterday, we can take this feature in a separate PR. Also, for this we might need to decide whether we want to look for the `.txt` files in only the base directory, or also in the sub-directories and provide user the option to select among the detected file(s).

transformer/classes/dockerfile/pythondockerfilegenerator.go
Akash-Nayak: Yes sure, done. I think same thing needs to be done in other transformers in the `dockerfile` folder too?

common/utils.go
```diff
@@ -91,6 +91,45 @@ func GetFilesByExt(inputPath string, exts []string) ([]string, error) {
...
+		}
+		return []string{}, nil
+	}
+	dirhandle, err := os.Open(inputPath)
+	if err != nil {
+		logrus.Warnf("Error in accessing %s due to %s", inputPath, err)
+		return []string{}, err
+	}
+	defer dirhandle.Close()
+	currDirFiles, err := dirhandle.Readdir(-1)
```
ashokponkumar, line 115: Won't using os.ReadDir be concise?

common/utils.go
```diff
@@ -91,6 +91,45 @@ func GetFilesByExt(inputPath string, exts []string) ([]string, error) {
...
+	if err != nil {
+		logrus.Warnf("Error while reading the directory: %s", err)
+		return []string{}, err
+	}
+	for _, currDirFile := range currDirFiles {
+		if currDirFile.Mode().IsRegular() {
```
ashokponkumar, line 121: With ReadDir, I think the Mode() function might not be required.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	var pythonMainFiles []string
+	for _, pythonFilePath := range pythonFilesPath {
+		pythonFile, err := ioutil.ReadFile(pythonFilePath)
+		if err != nil {
+			logrus.Warnf("Error in reading python file %s: %s.", pythonFilePath, err)
+		} else {
+			if strings.Contains(string(pythonFile), pythonMain1) || strings.Contains(string(pythonFile), pythonMain2) {
+				pythonMainFiles = append(pythonMainFiles, pythonFilePath)
+			}
```
ashokponkumar, line 90: @Akash-Nayak This is better done as a regex than contains. My previous comment on contains was only for django search in requirements file. 

For main check, it is important to use regex, because we also need to ensure it is at start of a line, and is more than one possibility for string and spaces. So a regex like say `^__name__\s*==\s*['"]__main__['"]\s*$ might capture most possibilities in a single check.

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		logrus.Errorf("Error while finding python files %s", err)
+		return nil, nil
+	}
+	if len(pythonFiles) == 0 {
+		return nil, nil
+	}
+	mainPythonFilesPath, err := findMainScripts(pythonFiles)
+	if err != nil {
+		return nil, err
```
ashokponkumar, line 132: Will python programs always need to have a main part? If there is no main should we ask users which is the starting python file in transform function from all the py files in the base project directory?

transformer/classes/dockerfile/pythondockerfilegenerator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	if len(mainPythonFilesPath) != 0 {
+		var serviceName, requirementsTxtPath string
+		var isDjangoProject bool
+		var detectedPorts []int32
+		detectedPorts = append(detectedPorts, 8080) //TODO: Write parser to parse and identify port
```
ashokponkumar, line 138: wouldn't it be better to initialize it like `detectedPorts = []int32{8080}`

transformer/classes/dockerfile/pythondockerfilegenerator.go
ashokponkumar: Actually remove the continue for now. We can consider this later.. Have to think through all scenarios, it might end up with error in other places, but that's fine. In most cases, this situation should not occur.

ashokponkumar (changes requested):
Nice work! Once this set of comments are addressed, I think we should be ready to merge.

ashokponkumar (approved):

ashokponkumar (approved):

PR #643, opened 2021-11-22
feat:  CustomTemplate path mapping to switch template-delimiter

seshapad:

### DIFF
--- a/filesystem/templatecopy.go
+++ b/filesystem/templatecopy.go
@@ -26,6 +26,15 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	DelimiterPairKey        = "delimiterPair"
+	ConfigKey               = "config"
+	DefaultOpeningDelimiter = "{{"
+	DefaultClosingDelimiter = "}}"
+	CustomOpeningDelimiter  = "[["
+	CustomClosingDelimiter  = "]]"
+)
+
 // TemplateCopy copies a directory to another and applies a template config on all files in the directory
 func TemplateCopy(source, destination string, config interface{}) error {
 	options := options{
@@ -38,7 +47,8 @@ func TemplateCopy(source, destination string, config interface{}) error {
 	return newProcessor(options).process(source, destination)
 }
 
-func templateCopyProcessFileCallBack(sourceFilePath, destinationFilePath string, config interface{}) error {
+func templateCopyProcessFileCallBack(sourceFilePath, destinationFilePath string, addOnConfig interface{}) error {
+	config := addOnConfig.(map[string]interface{})[ConfigKey]
 	si, err := os.Stat(sourceFilePath)
 	if err != nil {
 		logrus.Errorf("Unable to stat file %s : %s", sourceFilePath, err)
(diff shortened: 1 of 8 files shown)


filesystem/templatecopy.go
```diff
@@ -26,6 +26,21 @@ import (
...
+	// DelimiterPairKey is key for the delimeter pair in add-on config
+	DelimiterPairKey = "delimiterPair"
+	// ConfigKey is key for the template config in add-on config
+	ConfigKey = "config"
+	// DefaultOpeningDelimiter is default opening delimiter used in golang templates
+	DefaultOpeningDelimiter = "{{"
+	// DefaultClosingDelimiter is default closing delimiter used in golang templates
+	DefaultClosingDelimiter = "}}"
+	// CustomOpeningDelimiter is custom opening delimiter used in golang templates
+	CustomOpeningDelimiter = "[["
```
ashokponkumar, line 39: [[ is used in other context, so there is a chance it might conflict in many cases, like say startup command etc.. https://unix.stackexchange.com/questions/306111/what-is-the-difference-between-the-bash-operators-vs-vs-vs

filesystem/templatecopy.go
```diff
@@ -38,7 +53,8 @@ func TemplateCopy(source, destination string, config interface{}) error {
 	return newProcessor(options).process(source, destination)
 }
 
-func templateCopyProcessFileCallBack(sourceFilePath, destinationFilePath string, config interface{}) error {
+func templateCopyProcessFileCallBack(sourceFilePath, destinationFilePath string, addOnConfig interface{}) error {
+	config := addOnConfig.(map[string]interface{})[ConfigKey]
```
ashokponkumar, line 57: Would be good to define a type for this, so that it is explicit. We can use common.GetObjFromInterface to get the object from the interface.

types/transformer/pathmapping.go
```diff
@@ -30,6 +30,8 @@ const (
 	ModifiedSourcePathMappingType PathMappingType = "SourceDiff" // Source path becomes relative to source directory
 	// PathTemplatePathMappingType allows for path template registration
 	PathTemplatePathMappingType PathMappingType = "PathTemplate" // Path Template type
+	// CustomTemplatePathMappingType allows copy of source to destination and applying of template with custom delimiter
+	CustomTemplatePathMappingType PathMappingType = "CustomTemplate" // Source path when relative, is relative to yaml file location
```
ashokponkumar, line 34: Would Special be a better term here? Since it is not a customizable template, rather a special case to allow for specific usecases?

transformer/processors.go
```diff
@@ -64,7 +64,15 @@ func processPathMappings(pms []transformertypes.PathMapping, sourcePath, outputP
...
-			if err := filesystem.TemplateCopy(pm.SrcPath, destPath, pm.TemplateConfig); err != nil {
+			if err := filesystem.TemplateCopy(pm.SrcPath, destPath,
+				map[string]interface{}{filesystem.DelimiterPairKey: []string{filesystem.DefaultOpeningDelimiter, filesystem.DefaultClosingDelimiter},
+					filesystem.ConfigKey: pm.TemplateConfig}); err != nil {
+				logrus.Errorf("Error while copying sourcepath for %+v . Error: %q", pm, err)
+			}
+		case strings.ToLower(transformertypes.CustomTemplatePathMappingType):
+			if err := filesystem.TemplateCopy(pm.SrcPath, destPath,
+				map[string]interface{}{filesystem.DelimiterPairKey: []string{filesystem.CustomOpeningDelimiter, filesystem.CustomClosingDelimiter},
+					filesystem.ConfigKey: pm.TemplateConfig}); err != nil {
```
ashokponkumar, line 75: Explicit types might make it more readable.

ashokponkumar (changes requested):
Nice work! Just a few suggestions.

transformer/processors.go
seshapad: done

types/transformer/pathmapping.go
seshapad: done

filesystem/templatecopy.go
seshapad: done

filesystem/templatecopy.go
seshapad: done

ashokponkumar (approved):

PR #648, opened 2021-11-23
feat: Delete path-mapping to allow for file/folder deletion

seshapad:

### DIFF
--- a/transformer/processors.go
+++ b/transformer/processors.go
@@ -88,13 +88,18 @@ func processPathMappings(pms []transformertypes.PathMapping, sourcePath, outputP
 	}
 
 	for _, pm := range pms {
-		srcPath := pm.SrcPath
+		destPath := pm.DestPath
 		if !filepath.IsAbs(pm.DestPath) {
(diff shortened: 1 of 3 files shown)


transformer/processors.go
```diff
@@ -84,5 +86,21 @@ func processPathMappings(pms []transformertypes.PathMapping, sourcePath, outputP
...
 		}
 	}
+
+	for _, pm := range pms {
+		destPath := pm.DestPath
+		if !filepath.IsAbs(pm.DestPath) {
+			destPath = filepath.Join(outputPath, pm.DestPath)
+		}
+		switch strings.ToLower(pm.Type) {
+		case strings.ToLower(transformertypes.DeletePathMappingType):
```
ashokponkumar, line 96: Do we need a switch case for this? Why not just a if with Equalsfold?

ashokponkumar (changes requested):
Nice work @seshapad . Just a small suggestion.

transformer/processors.go
seshapad: done.

ashokponkumar (approved):

PR #651, opened 2021-11-26
feat: Set umask only in non-windows platforms

ashokponkumar:

### DIFF
--- a/cmd/move2kube/umask.go
+++ b/cmd/move2kube/umask.go
@@ -1,0 +1,31 @@
+//go:build !windows
+// +build !windows
+
+/*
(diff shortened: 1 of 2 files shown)


cmd/move2kube/umask.go
```diff
@@ -0,0 +1,31 @@
+//go:build !windows
+// +build !windows
```
HarikrishnanBalagopal, line 2: what does the 2nd line do?

HarikrishnanBalagopal (commented):
Isn't the first line enough?

HarikrishnanBalagopal (approved):
LGTM

PR #666, opened 2021-12-03
chore: update the screenshots with the new UI

HarikrishnanBalagopal:


ashokponkumar:
Thanks Hari! Very glad to have it.

One question I have is the number of images we are uploading, and especially since this is our main repo. Is there a way to keep the images in some other location, other than the main repo, so that when we change/add new images, we don't increase the size of our main repo?

HarikrishnanBalagopal:
> Thanks Hari! Very glad to have it.
> 
> One question I have is the number of images we are uploading, and especially since this is our main repo. Is there a way to keep the images in some other location, other than the main repo, so that when we change/add new images, we don't increase the size of our main repo?

yes we can do that. I am not sure where we could upload it though. Are you thinking of making a new repo like `konveyor/move2kube-website-images` just for the images?

For now, I have compressed the images so that they are all less 100KB in size.

docs/tutorials/ui-local-workflow.md
```diff
@@ -10,315 +10,298 @@ nav_order: 4
...
-   LICENSE            code-of-conduct.md docker-compose.yml server.js          webpack.common.js  wksps
-   Makefile           containers.txt     dr-surge.js        src                webpack.dev.js     yarn.lock
-   ```
-3. Download [language-platforms.zip](https://github.com/konveyor/move2kube-demos/raw/main/samples/language-platforms.zip) which we will be using for this tutorial. The language-platforms.zip file has a combination of multiple applications in different languages (Java, Go, Python, Ruby, etc.) which needs to be containerized and then put into Kubernetes.
+1. Download [language-platforms.zip](https://github.com/konveyor/move2kube-demos/raw/main/samples/language-platforms.zip) which we will be using for this tutorial. The language-platforms.zip file has a combination of multiple applications in different languages (Java, Go, Python, Ruby, etc.) which needs to be containerized and then put into Kubernetes.
 
 ## Steps to generate target artifacts
 
-1. Do a `docker-compose up`.
+1. Do a `docker run --rm -it -p 8080:8080 quay.io/konveyor/move2kube-ui`.
```
ashokponkumar, line 23: Would be good to have the command with mounting

ashokponkumar (changes requested):
Nice! Just a small suggestion. It would be good to get our users accustomed to persistent storage based usage.

docs/tutorials/ui-local-workflow.md
HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #693, opened 2021-12-17
feat: Add netcoreapp3.1 and net6.0 to the supported dotnetcore frameworks

Akash-Nayak:


transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
```diff
@@ -39,7 +39,7 @@ import (
 
 var (
 	portRegex            = regexp.MustCompile(`:(\d+)`)
-	dotNetCoreFrameworks = []string{"net5.0", "netcoreapp2.1", "netstandard2.0"}
+	dotNetCoreFrameworks = []string{"net5.0", "netcoreapp2.1", "netstandard2.0", "netcoreapp3.1"}
```
ashokponkumar, line 42: @Akash-Nayak As discussed, lets try to use a regex or some other way so that we can cover the full spectrum of versions, rather than trying to support each version individually.

ashokponkumar (changes requested):

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
```diff
@@ -128,7 +129,18 @@ func getDotNetCoreVersion(mappingFile string, name string) (version string, err
...
+		dotNetCoreFramework = dotnetcoreRegex1.FindString(name)
+	} else if dotnetcoreRegex2.MatchString(name) {
+		dotNetCoreFramework = dotnetcoreRegex2.FindString(name)
+	} else if dotnetcoreRegex3.MatchString(name) {
+		dotNetCoreFramework = dotnetcoreRegex3.FindString(name)
+	}
+	if _, ok := dotNetCoreNameVersionMapping.Spec.NameVersion[dotNetCoreFramework]; ok {
+		return dotNetCoreNameVersionMapping.Spec.NameVersion[dotNetCoreFramework], nil
+	}
+	return "", fmt.Errorf("could not find compatible .NET Core framework version")
```
ashokponkumar, line 143: @Akash-Nayak the goal of the regex is to avoid the explicit version mappings requirements for all possible combination. We need to use some sort of heuristic to determine the best mapping. 

If the version numbers at the end of the name is always going to be the image name, and it is always supposed to be present, why not just get that value alone through regex and use that as the image name?

ashokponkumar (changes requested):

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
Akash-Nayak: Yes, I was also thinking that we should either have the version mappings or the regex-based heuristic. I have removed the version mappings now.

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
```diff
@@ -38,8 +37,11 @@ import (
 )
 
 var (
-	portRegex            = regexp.MustCompile(`:(\d+)`)
-	dotNetCoreFrameworks = []string{"net5.0", "netcoreapp2.1", "netstandard2.0"}
+	portRegex        = regexp.MustCompile(`:(\d+)`)
+	versionRegex     = regexp.MustCompile(`\d+\.\d+`)
+	dotnetcoreRegex1 = regexp.MustCompile(`net\d+\.\d+`)
+	dotnetcoreRegex2 = regexp.MustCompile(`netcoreapp\d+\.\d+`)
+	dotnetcoreRegex3 = regexp.MustCompile(`netstandard\d+\.\d+`)
```
ashokponkumar, line 44: Do we need 4 regex here? Can we use something like `net(?:(?:coreapp)|(?:standard))?(\d+\.\d+)` , and use the capturing group to get the version with just a single regex?

ashokponkumar (changes requested):

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
Akash-Nayak: Yes sure, thank you. I have updated the regex.

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
```diff
@@ -420,8 +398,11 @@ func (t *DotNetCoreDockerfileGenerator) Transform(newArtifacts []transformertype
 			logrus.Errorf("Could not read the project file (%s) : %s", filepath.Join(a.Paths[artifacts.ProjectPathPathType][0], csprojRelFilePath), err)
 			continue
 		}
-		dotNetCoreTemplateConfig.DotNetVersion, err = getDotNetCoreVersion(filepath.Join(t.Env.GetEnvironmentContext(), "mappings/dotnetcoreversions.yaml"), csprojConfiguration.PropertyGroup.TargetFramework)
-		if err != nil || dotNetCoreTemplateConfig.DotNetVersion == "" {
+		frameworkVersion := dotnetcoreRegex.FindAllStringSubmatch(csprojConfiguration.PropertyGroup.TargetFramework, -1)
+		if len(frameworkVersion[0]) == 2 {
```
ashokponkumar, line 402: This will panic if framework version is an empty array. Wouldn't it be better to check is length too?

ashokponkumar (changes requested):

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
Akash-Nayak: Yes, I was just updating it. Pushed the changes now.

ashokponkumar (approved):

PR #706, opened 2022-01-03
fix: Format the interfaces in map values as json

seshapad:

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -1072,8 +1072,8 @@ func StringifyMap(inputMap map[string]interface{}) map[string]interface{} {
 			inputMap[key] = nil
 			continue
 		}
-		if reflect.TypeOf(value).Kind() == reflect.String {
-			inputMap[key] = value.(string)
+		if val, ok := value.(string); ok {
+			inputMap[key] = val
 			continue
 		}
 		var b bytes.Buffer
--- a/common/utils.go
+++ b/common/utils.go
@@ -1065,6 +1065,30 @@ func MarshalObjToYaml(obj runtime.Object) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+// StringifyMap stringifies the map values
+func StringifyMap(inputMap map[string]interface{}) map[string]interface{} {
+	for key, value := range inputMap {
+		if value == nil {
+			inputMap[key] = nil
+			continue
+		}
+		var b bytes.Buffer
+		encoder := json.NewEncoder(&b)
+		if err := encoder.Encode(value); err != nil {
+			logrus.Error("Unable to unmarshal data to json while converting map interfaces to string")
+			continue
+		}
+		strValue := string(b.Bytes())
+		strValue = strings.TrimSpace(strValue)
+		strValue = strings.Trim(strValue, `"\r\n`)
+		if strValue == "{}" {
+			strValue = ""
+		}
+		inputMap[key] = strValue
+	}
+	return inputMap
+}
+
 // ConvertInterfaceToSliceOfStrings converts an interface{} to a []string type.
 // It can handle []interface{} as long as all the values are strings.
 func ConvertInterfaceToSliceOfStrings(xI interface{}) ([]string, error) {
--- a/common/utils.go
+++ b/common/utils.go
@@ -1072,6 +1072,10 @@ func StringifyMap(inputMap map[string]interface{}) map[string]interface{} {
 			inputMap[key] = nil
 			continue
 		}
+		if reflect.TypeOf(value).Kind() == reflect.String {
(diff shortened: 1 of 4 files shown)


common/utils.go
```diff
@@ -1065,6 +1065,31 @@ func MarshalObjToYaml(obj runtime.Object) ([]byte, error) {
...
+		}
+		var b bytes.Buffer
+		encoder := json.NewEncoder(&b)
+		if err := encoder.Encode(value); err != nil {
+			logrus.Error("Unable to unmarshal data to json while converting map interfaces to string")
+			continue
+		}
+		strValue := string(b.Bytes())
+		strValue = strings.TrimSpace(strValue)
+		strValue = strings.Trim(strValue, `\r\n`)
```
ashokponkumar, line 1087: Wont trimspace remove this too?

common/utils.go
```diff
@@ -1065,6 +1065,31 @@ func MarshalObjToYaml(obj runtime.Object) ([]byte, error) {
...
 
+// StringifyMap stringifies the map values
+func StringifyMap(inputMap map[string]interface{}) map[string]interface{} {
+	for key, value := range inputMap {
+		if value == nil {
+			inputMap[key] = nil
+			continue
+		}
+		if reflect.TypeOf(value).Kind() == reflect.String {
+			inputMap[key] = value.(string)
```
ashokponkumar, line 1076:
```suggestion
		if val, ok := value.(string); ok {
			inputMap[key] = val
```

common/utils.go
```diff
@@ -1065,6 +1065,31 @@ func MarshalObjToYaml(obj runtime.Object) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+// StringifyMap stringifies the map values
+func StringifyMap(inputMap map[string]interface{}) map[string]interface{} {
+	for key, value := range inputMap {
+		if value == nil {
+			inputMap[key] = nil
```
ashokponkumar, line 1072: do we need this assignment if it is already nil?

common/utils.go
```diff
@@ -1065,6 +1065,31 @@ func MarshalObjToYaml(obj runtime.Object) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+// StringifyMap stringifies the map values
+func StringifyMap(inputMap map[string]interface{}) map[string]interface{} {
```
ashokponkumar, line 1069: StringifyMap is a very specify function that is not a general usecase, so won't it be better to just place it in the collector or collect type package itself? or do we see this specific situation occurring in other places?

transformer/cloudfoundrytransformer.go
```diff
@@ -321,6 +321,7 @@ func getCfAppInstance(path string, appname string) (collecttypes.CfApp, error) {
 	if err := common.ReadMove2KubeYaml(path, &cfinstanceappsfile); err != nil {
 		return collecttypes.CfApp{}, err
 	}
+	cfinstanceappsfile = collecttypes.FormatMapsWithInterface(cfinstanceappsfile)
```
ashokponkumar, line 324: Do we need to call this here? This will anyway be loaded as a string, isn't it?

ashokponkumar (changes requested):
Nice. Few comments..

common/utils.go
seshapad: done

common/utils.go
seshapad: yes. made changes accordingly.

common/utils.go
seshapad: done.

transformer/cloudfoundrytransformer.go
seshapad: done.

seshapad (commented):

types/collection/cfapps.go
```diff
@@ -52,3 +57,42 @@ func NewCfApps() CfApps {
...
+		}
+		var b bytes.Buffer
+		encoder := json.NewEncoder(&b)
+		if err := encoder.Encode(value); err != nil {
+			logrus.Error("Unable to unmarshal data to json while converting map interfaces to string")
+			continue
+		}
+		strValue := string(b.Bytes())
+		strValue = strings.TrimSpace(strValue)
+		//strValue = strings.Trim(strValue, `\r\n`)
```
ashokponkumar, line 94: So we need this line?

ashokponkumar (changes requested):
Minor comment. Also, did we check with pr with e2e app?

seshapad:
> Minor comment. Also, did we check with pr with e2e app?

@ashokponkumar yes. It works!

ashokponkumar (approved):

PR #707, opened 2022-01-03
docs: expose all the commands to help generate docs externally

HarikrishnanBalagopal:
Expose commands in a package other than main so they can be imported.
Also add a hidden command to generate the documentation.


cmd/commands/docs.go
```diff
@@ -0,0 +1,107 @@
...
+	viper.AutomaticEnv()
+	flags := generateFlags{}
+	generateDocsCmd := &cobra.Command{
+		Hidden: true,
+		Use:    "docs",
+		Short:  "Generate the documentation for the commands",
+		Long:   "Generate the documentation for the commands. The documentation is in markdown format.",
+		Run:    func(_ *cobra.Command, __ []string) { generateHandler(flags) },
+	}
+	generateDocsCmd.Flags().StringVarP(&flags.outputDir, "output", "o", "commands", "path to the directory where the documentation will be generated")
```
ashokponkumar, line 105:
```suggestion
	generateDocsCmd.Flags().StringVarP(&flags.outputDir, "output", "o", "commands", "Path to the directory where the documentation will be generated.")
```

ashokponkumar (changes requested):
Nice! We have tried to avoid hardcoding `move2kube` in strings in most places, and use AppName instead. Would be good to follow that here too.

cmd/commands/docs.go
```diff
@@ -0,0 +1,107 @@
...
+package commands
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	"github.com/konveyor/move2kube/common"
```
ashokponkumar, line 26:
```suggestion
	"github.com/konveyor/move2kube/common"
	"github.com/konveyor/move2kube/types"
```

ashokponkumar (changes requested):

ashokponkumar (approved):

Issue #685, opened 2021-12-14
The m2k_collect cluster output contains `|` in the metadata name


Akash-Nayak:
**Describe the bug**
The m2k_collect cluster output has `|` in the metadata name and the actual name is in the next line, due to which the plan file has incorrect name of custom cluster type in the `targetClusters`.

**To Reproduce**
1. Login to the cluster and run move2kube collect.
2. Check the output cluster yaml.
3. Run `move2kube plan -s /path/to/src/ -c m2k_collect/clusters/`.
4. Check the m2k.plan file.

Example:
cluster yaml
```
apiVersion: move2kube.konveyor.io/v1alpha1
kind: ClusterMetadata
metadata:
  name: |
    cf2ocp/c114-e-us-south-containers:32230/IAM#xyz@gmail.com

```

m2.plan file
```
    targetClusters:
      Kubernetes: m2kassets/inbuilt/clusters/kubernetes.yaml
      Openshift: m2kassets/inbuilt/clusters/openshift.yaml
      ? |
        cf2ocp/c114-e-us-south-containers:32230/IAM#xyz@gmail.com
      : m2kassets/custom/IAM-xyz-1f50a05161df3b57.yaml
```

**Expected behavior**
cluster yaml metadata should not contain `|`
```
apiVersion: move2kube.konveyor.io/v1alpha1
kind: ClusterMetadata
metadata:
  name: cf2ocp/c114-e-us-south-containers:32230/IAM#xyz@gmail.com

```

m2.plan file
```
    targetClusters:
      Kubernetes: m2kassets/inbuilt/clusters/kubernetes.yaml
      Openshift: m2kassets/inbuilt/clusters/openshift.yaml
      cf2ocp/c114-e-us-south-containers:32230/IAM#xyz@gmail.com: m2kassets/custom/IAM-xyz-1f50a05161df3b57.yaml
```

**Desktop (please complete the following information):**
 - OS: macOS
 - Version: v0.3.0+unreleased
 - gitCommit: dcf8793a889c0a8f9f4423e9e9ee3a95003c6bcc

HarikrishnanBalagopal:
Might be related to this https://github.com/konveyor/move2kube/issues/178

PR #709, opened 2022-01-04
fix: Remove space from cluster name in move2kube collect

ashokponkumar:


ashokponkumar:
Fixes #685

PR #711, opened 2022-01-04
feat: check the directory the script is being run from and give a warning

HarikrishnanBalagopal:


ashokponkumar:
Was there an issue with the approach 1 of using $0?

ashokponkumar (approved):

PR #725, opened 2022-01-12
feat: Created a secret for all Cf VCAP environments

seshapad:


common/constants.go
```diff
@@ -180,6 +180,10 @@ const (
 	VcapServiceEnvName = "VCAP_SERVICES"
 	// VcapApplicationEnvName refers to the VCAP_APPLICATION environment variable DefaultDockerfileName
 	VcapApplicationEnvName = "VCAP_APPLICATION"
+	// VcapSpringBootSecretSuffix refers to VCAP springboot secret suffix
+	VcapSpringBootSecretSuffix = "-vcapspringboot"
+	// VcapCfSecretSuffix refers to VCAP secret suffix
+	VcapCfSecretSuffix = "-vcapcf"
```
ashokponkumar, line 186: can we call it something like say, -vcapasspringbootproperties and -vcapasenv? also possibly change the description to state what they actually store.

transformer/cloudfoundrytransformer.go
```diff
@@ -244,8 +250,38 @@ func (t *CloudFoundry) Transform(newArtifacts []transformertypes.Artifact, alrea
 	return nil, artifactsCreated, nil
 }
 
+// NormalizeForEnvironmentVariableName converts the string to be compatible for environment variable name
+func (t *CloudFoundry) normalizeForEnvironmentVariableName(envName string) string {
+	newName := common.DisallowedDNSCharactersRegex.ReplaceAllLiteralString(strings.ToLower(envName), "_")
```
ashokponkumar, line 255: the env variable need for follow DNS regex. Lets create a new regex with things allowed in a env variable. Also use the regex in a way that we don't need to use tolower and immediately follow it by toupper. Also we can move this function to common pacakge

transformer/cloudfoundrytransformer.go
```diff
@@ -244,8 +250,38 @@ func (t *CloudFoundry) Transform(newArtifacts []transformertypes.Artifact, alrea
...
+		}
+		env.Name = newEnvName
+		envList[index] = env
+	}
+	for key, value := range vcapMap {
+		newKey := t.normalizeForEnvironmentVariableName(key)
+		newVcapMap[newKey] = value
+	}
+	return envList, newVcapMap
+}
```
ashokponkumar, line 279: Do we need this function? Why not call normalizeForEnvironmentVariableName directly befoer 301?

transformer/cloudfoundrytransformer.go
```diff
@@ -263,21 +299,43 @@ func (t *CloudFoundry) prioritizeAndAddEnvironmentVariables(cfApp collecttypes.C
...
+				vcapEnvMap[env.Name] = []byte(env.Value)
 			}
 		}
+		dbEnv, err := createDbEnvironmentVariable(valueStr, serviceName)
+		if err != nil {
+			logrus.Errorf("Could not create DATABASE_URL environment: %v", err)
+		} else {
+			envOrderMap[dbEnv.Name] = dbEnv
+			vcapEnvMap[dbEnv.Name] = []byte(dbEnv.Value)
+		}
```
ashokponkumar, line 311: shouldn't this be done within the if condition above?

transformer/cloudfoundrytransformer.go
```diff
@@ -263,21 +299,43 @@ func (t *CloudFoundry) prioritizeAndAddEnvironmentVariables(cfApp collecttypes.C
...
 		envOrderMap[varname] = core.EnvVar{Name: varname, Value: valueStr}
+		vcapEnvMap[varname] = []byte(valueStr)
```
ashokponkumar, line 313: why would this need to be in vcapenvmap?

transformer/cloudfoundrytransformer.go
```diff
@@ -263,21 +299,43 @@ func (t *CloudFoundry) prioritizeAndAddEnvironmentVariables(cfApp collecttypes.C
 			flattenedEnvList := flattenVcapServiceVariables(valueStr)
 			for _, env := range flattenedEnvList {
 				envOrderMap[env.Name] = env
```
ashokponkumar, line 301: Why not call normalizeForEnvironmentVariableName function directly here?

transformer/cloudfoundrytransformer.go
```diff
@@ -263,21 +299,43 @@ func (t *CloudFoundry) prioritizeAndAddEnvironmentVariables(cfApp collecttypes.C
...
+		} else {
+			envOrderMap[dbEnv.Name] = dbEnv
+			vcapEnvMap[dbEnv.Name] = []byte(dbEnv.Value)
+		}
 		envOrderMap[varname] = core.EnvVar{Name: varname, Value: valueStr}
+		vcapEnvMap[varname] = []byte(valueStr)
 	}
 	for varname, value := range cfApp.Environment.ApplicationEnv {
-		envOrderMap[varname] = core.EnvVar{Name: varname, Value: fmt.Sprintf("%s", value)}
+		valueStr := fmt.Sprintf("%s", value)
```
ashokponkumar, line 316: Why not call normalizeForEnvironmentVariableName function directly here?

transformer/cloudfoundrytransformer.go
```diff
@@ -301,13 +359,34 @@ func flattenVariable(prefix string, credential interface{}) []core.EnvVar {
...
+			if serviceInstance.ServiceName == serviceName {
+				if uriValue, ok := serviceInstance.ServiceCredentials["uri"].(string); ok {
+					return core.EnvVar{Name: "DATABASE_URL", Value: uriValue}, nil
+				}
+				return core.EnvVar{}, fmt.Errorf("uri field is not available in service credential for the service VCAP_SERVICES")
+			}
+		}
+	}
+	return core.EnvVar{}, fmt.Errorf("service %s not available in VCAP_SERVICES", serviceName)
+}
```
ashokponkumar, line 381: Do we need a separate function for this? why not do it within flattenVcapServiceVariables? We are already casting it into a vcap object there and also iterating over the services there.

ashokponkumar (changes requested):
Few suggestions...

common/constants.go
seshapad: done

transformer/cloudfoundrytransformer.go
seshapad: done

transformer/cloudfoundrytransformer.go
seshapad: Because vcapenvmap contains all variables which go into the secret. This includes the vcap flattened environment variables as well as the ones like DATABASE_URL generated using vcap values.

transformer/cloudfoundrytransformer.go
seshapad: done

transformer/cloudfoundrytransformer.go
seshapad: we strictly dont need normalizeForEnvironmentVariableName() in this location as we are not unpacking VCAP_APPLICATION variable.

transformer/cloudfoundrytransformer.go
seshapad: done

transformer/cloudfoundrytransformer.go
seshapad: This comment is moot. I have removed createDbEnvironmentVariable().

transformer/cloudfoundrytransformer.go
seshapad: Need some more clarification on this.

transformer/cloudfoundrytransformer.go
seshapad: done based on clarification.

common/constants.go
```diff
@@ -202,8 +206,10 @@ var (
 	DisableLocalExecution = false
 	// DefaultIgnoreDirRegexps specifies directory name regexes that would be ignored
 	DefaultIgnoreDirRegexps = []*regexp.Regexp{regexp.MustCompile("^[.].*")}
-	// Characters not allowed in a DNS Name
-	disallowedDNSCharactersRegex = regexp.MustCompile(`[^a-z0-9\-]`)
+	// DisallowedDNSCharactersRegex provides pattern for characters not allowed in a DNS Name
+	DisallowedDNSCharactersRegex = regexp.MustCompile(`[^a-z0-9\-]`)
```
ashokponkumar, line 210: Do we need to expose it outside the package?

common/utils.go
```diff
@@ -497,13 +497,23 @@ func NormalizeForFilename(name string) string {
...
 	if newName != svcName {
 		logrus.Infof("Changing service name to %s from %s", svcName, newName)
 	}
 	return newName
 }
 
+// NormalizeForEnvironmentVariableName converts the string to be compatible for environment variable name convention specified below:
+// https://pubs.opengroup.org/onlinepubs/9699919799/
+func NormalizeForEnvironmentVariableName(envName string) string {
+	newName := DisallowedEnvironmentCharactersRegex.ReplaceAllLiteralString(envName, "_")
```
ashokponkumar, line 510: This won't work. This will replace only if the first character is a digit

common/constants.go
```diff
@@ -202,8 +206,10 @@ var (
...
+	// DisallowedEnvironmentCharactersRegex provides pattern for characters not allowed in a DNS Name
+	DisallowedEnvironmentCharactersRegex = regexp.MustCompile(`^[^A-Z]|[^A-Z0-9\_]`)
```
ashokponkumar, line 212: Do we need to expose this variable outside the package

ashokponkumar (changes requested):

ashokponkumar (approved):
Perfect. Thanks!

PR #738, opened 2022-01-17
fix: install correct java version and liberty version in the runner image

HarikrishnanBalagopal:


ashokponkumar (commented):
We have to change the liberty transformer implementation too to not go for liberty images and removing the mapping file.

ashokponkumar (approved):

ashokponkumar (approved):

PR #740, opened 2022-01-24
fix: Update the download link of Tomcat in Dockerfile template

Akash-Nayak:


ashokponkumar (changes requested):
Nice catch! Since we are updating this, lets update the tomcat version also to the latest.

Akash-Nayak:
> Nice catch! Since we are updating this, lets update the tomcat version also to the latest.

Done.

ashokponkumar (approved):

PR #748, opened 2022-02-01
fix: prevent cf transformer creation of ir when containerizing app

ashokponkumar:


HarikrishnanBalagopal (approved):
LGTM

PR #749, opened 2022-02-01
fix: remove assumption of port 8080 for cf apps when no port is found

ashokponkumar:


HarikrishnanBalagopal (approved):
LGTM

PR #761, opened 2022-02-08
feat: Allow clusterip only service

ashokponkumar:


HarikrishnanBalagopal (approved):
LGTM, we will look at changing this to a multiple choice question in a later PR

PR #785, opened 2022-03-14
fix: Normalize the git repo name for the service name

Akash-Nayak:


ashokponkumar (changes requested):
Don't we need to do this in 106 and 99 lines too?

Akash-Nayak:
> Don't we need to do this in 106 and 99 lines too?

Done.

ashokponkumar (approved):

PR #791, opened 2022-03-24
fix: allow specifying the environments to parameterize for

HarikrishnanBalagopal:


ashokponkumar:
How do users specify to have only values yaml without any env?

HarikrishnanBalagopal:
> How do users specify to have only values yaml without any env?

```
envs:
  - default
```

HarikrishnanBalagopal:
I have changed it to move`dev`, `staging` and `prod` defaults to the config section in the `transformer.yaml` file.

ashokponkumar (approved):

PR #798, opened 2022-05-02
feat: log the transformer names and artifacts to a graph

HarikrishnanBalagopal:

image


ashokponkumar (changes requested):
Wow! Nice work @HarikrishnanBalagopal . Will greatly improve debuggability of Move2Kube. 

Few feature suggestions Hari. We can do it in another PR.

a. We should add this to Move2Kube ui too.
b. Would be good to make it a live graph, where the transform command continuously outputs the json, and the ui keeps refreshing.
c. If it is a live graph, we could show the current active node.
d. On hover, we can show the actual contents of the artifacts.

Couple of questions:
a. Why is readmegenerator and parameterizer listed on the top? Is it yet to be executed?
b. There seems to be repetitions of parameterizer and readmegenerator.
c. We are showing dependency in the same level as the iteration, which might be confusing. We have to check how we handle dependency and mandatory pass-through.

HarikrishnanBalagopal:
> Wow! Nice work @HarikrishnanBalagopal . Will greatly improve debuggability of Move2Kube.
> 
> Few feature suggestions Hari. We can do it in another PR.
> 
> a. We should add this to Move2Kube ui too. b. Would be good to make it a live graph, where the transform command continuously outputs the json, and the ui keeps refreshing. c. If it is a live graph, we could show the current active node. d. On hover, we can show the actual contents of the artifacts.
> 
> Couple of questions: a. Why is readmegenerator and parameterizer listed on the top? Is it yet to be executed? b. There seems to be repetitions of parameterizer and readmegenerator. c. We are showing dependency in the same level as the iteration, which might be confusing. We have to check how we handle dependency and mandatory pass-through.

- changed the block positions to match iterations (taking into account mandatory/on-demand passthrough)
- added path mappings
image

ashokponkumar (approved):
Nice work!

PR #799, opened 2022-05-04
fix: simplify the build process by removing the dependency on node and yarn. Check in the built web server directory so it can be embedded by Golang

HarikrishnanBalagopal:


ashokponkumar (approved):
Looks good.

PR #803, opened 2022-05-10
feat: add an ArgoCD transformer to generate the ArgoCD specific deployment yamls

HarikrishnanBalagopal:
fix: update dependencies to fix build errors after adding argocd
    
    Remove the useless -extldflags "-static" flag because that only applies
    to the external linker `ld`. Golang uses the internal Go compiler and linker
    which by default links everything statically.
    
    The external compiler `clang` and external linker `ld` only gets invoked
    when there is C code to be compiled. ArgoCD has some C code that needs to be
    compiled.
    
    Unfortunately there is no way to compile statically against system libraries
    on most systems (except Linux) so the -extldflags "-static" flag causes a build
    failure. So we need to remove that flag.
    
    Another alternative to removing the -extldflags "-static" flag is to disable CGO
    CGO_ENBALED=0 go build ..... does work but disables CGO for all packages. This
    will only work if all packages provide Golang replacements for the C code that they
    were using.
    
    This could cause panics/issues down the line if the Golang replacements provided by
    the various packages we use are mere stubs. It also will cause compilation failure
    if such replacements are not available for a particular package.
    
    There are more fine-grained ways to disable CGO for a single package like
    https://pkg.go.dev/os/user for example.
    
    "When cgo is available, and the required routines are implemented in libc
    for a particular platform, cgo-based (libc-backed) code is used. This can be overridden
    by using osusergo build tag, which enforces the pure Go implementation."
    
    See:
    https://www.reddit.com/r/golang/comments/53nl4b/why_all_go_binaries_are_staticallylinked/
    https://lists.sr.ht/~sircmpwn/public-inbox/%3C3A12328E-5EC6-4E34-A676-64A3039A2AF7%40herrbischoff.com%3E
    https://stackoverflow.com/questions/3801011/ld-library-not-found-for-lcrt0-o-on-osx-10-6-with-gcc-clang-static-flag
    https://stackoverflow.com/questions/844819/how-to-static-link-on-os-x
    

ashokponkumar:
Do we happen to know which library requires dynamic linking?

HarikrishnanBalagopal:
> Do we happen to know which library requires dynamic linking?

```
This option will not work on Mac OS X unless all libraries (including libgcc.a) have also been compiled with -static. Since neither a static version of libSystem.dylib nor crt0.o are provided, this option is not useful to most people.
```

```
$ otool -L move2kube 
move2kube:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1856.105.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 60157.60.19)
```

transformer/kubernetes/argocdtransformer.go
```diff
@@ -0,0 +1,155 @@
...
+	ArgoCDConfig *ArgoCDYamlConfig
+}
+
+// ArgoCDYamlConfig stores the ArgoCD related information
+type ArgoCDYamlConfig struct {
+	OutputPath string `yaml:"outputPath"`
+}
+
+const (
+	defaultArgoCDYamlsOutputPath = common.DeployDir + string(os.PathSeparator) + common.CICDDir + string(os.PathSeparator) + "argocd"
```
ashokponkumar, line 48: Why not use filepath.join?

transformer/kubernetes/argocdtransformer.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	return pathMappings, createdArtifacts, nil
+}
+
+// setupEnhancedIR returns EnhancedIR containing ArgoCD components
+func (t *ArgoCD) setupEnhancedIR(oldir irtypes.IR, projectName string) irtypes.EnhancedIR {
+	ir := irtypes.NewEnhancedIRFromIR(oldir)
+	// Prefix the project name and make the name a valid k8s name.
+	p := func(baseName string) string {
```
ashokponkumar, line 147: Why do we need a function for this?

transformer/kubernetes/buildconfigtransformer.go
```diff
@@ -60,13 +49,23 @@ type BuildConfigYamlConfig struct {
 	OutputPath string `yaml:"outputPath"`
 }
 
+const (
+	defaultBuildConfigYamlsOutputPath = common.DeployDir + string(os.PathSeparator) + common.CICDDir + string(os.PathSeparator) + "buildconfig"
```
ashokponkumar, line 53: lets use filepath.Join

ashokponkumar (changes requested):
Nice work! A few comments.

transformer/kubernetes/argocdtransformer.go
HarikrishnanBalagopal: because then it can't be a `const`

transformer/kubernetes/argocdtransformer.go
HarikrishnanBalagopal: I made it similar to Tekton. ArgoCD has other resources like `Project` which we may need to generate in the future.

transformer/kubernetes/buildconfigtransformer.go
HarikrishnanBalagopal: if we use a function then we have to change it to `var`.

ashokponkumar (approved):

HarikrishnanBalagopal:
Also see https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes#Kernel
```
New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)
```

PR #808, opened 2022-05-16
feat: Added default values for cli inputs

seshapad:
- source input
- customization

### DIFF
--- a/cmd/transform.go
+++ b/cmd/transform.go
@@ -94,9 +94,6 @@ func transformHandler(cmd *cobra.Command, flags transformFlags) {
 		if cmd.Flags().Changed(planFlag) {
 			logrus.Fatalf("Error while accessing plan file at path %s Error: %q", flags.planfile, err)
 		}
-		if !cmd.Flags().Changed(sourceFlag) {
-			logrus.Fatalf("Invalid usage. Must specify either path to a plan file or path to directory containing source code.")
-		}
 
 		// Global settings
 		checkSourcePath(flags.srcpath)
@@ -110,6 +107,9 @@ func transformHandler(cmd *cobra.Command, flags transformFlags) {
 		}
 		startQA(flags.qaflags)
 		logrus.Debugf("Creating a new plan.")
+		if _, err := os.Stat(flags.customizationsPath); os.IsNotExist(err) {
+			flags.customizationsPath = ""
+		}
 		p = lib.CreatePlan(ctx, flags.srcpath, flags.outpath, flags.customizationsPath, flags.transformerSelector, flags.name)
 	} else {
 		logrus.Infof("Detected a plan file at path %s. Will transform using this plan.", flags.planfile)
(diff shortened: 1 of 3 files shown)


cmd/transform.go
```diff
@@ -173,7 +173,7 @@ func GetTransformCommand() *cobra.Command {
 	// Basic options
 	transformCmd.Flags().StringVarP(&flags.planfile, planFlag, "p", common.DefaultPlanFile, "Specify a plan file to execute.")
 	transformCmd.Flags().BoolVar(&flags.overwrite, overwriteFlag, false, "Overwrite the output directory if it exists. By default we don't overwrite.")
-	transformCmd.Flags().StringVarP(&flags.srcpath, sourceFlag, "s", "", "Specify source directory to transform. If you already have a m2k.plan then this will override the sourceDir value specified in that plan.")
+	transformCmd.Flags().StringVarP(&flags.srcpath, sourceFlag, "s", common.DefaultSourceFolder, "Specify source directory to transform. If you already have a m2k.plan then this will override the sourceDir value specified in that plan.")
```
ashokponkumar, line 176: We should not do this. This will prevent the source directory in plan from ever activating.

cmd/transform.go
```diff
@@ -94,9 +94,6 @@ func transformHandler(cmd *cobra.Command, flags transformFlags) {
 		if cmd.Flags().Changed(planFlag) {
 			logrus.Fatalf("Error while accessing plan file at path %s Error: %q", flags.planfile, err)
 		}
-		if !cmd.Flags().Changed(sourceFlag) {
-			logrus.Fatalf("Invalid usage. Must specify either path to a plan file or path to directory containing source code.")
-		}
```
ashokponkumar, line 99: I think it is okay to expect the source directory to be specified and being mandatory.

cmd/transform.go
```diff
@@ -110,6 +107,9 @@ func transformHandler(cmd *cobra.Command, flags transformFlags) {
 		}
 		startQA(flags.qaflags)
 		logrus.Debugf("Creating a new plan.")
+		if _, err := os.Stat(flags.customizationsPath); os.IsNotExist(err) {
+			flags.customizationsPath = ""
+		}
```
ashokponkumar, line 112: This should happen only if the directory is default. Else it should be an error.

common/pathsofinterest.go
```diff
@@ -25,6 +25,10 @@ import (
 const (
 	// DefaultPlanFile defines default name for plan file
 	DefaultPlanFile = types.AppNameShort + ".plan"
+	// DefaultSourceFolder defines default name for plan file
+	DefaultSourceFolder = types.AppNameShort + "-input"
```
ashokponkumar, line 29: Lets not have a default for source, it might not be obvious

ashokponkumar (changes requested):
Nice work @seshapad! We should also possibly add default for config file too?

seshapad:
> Nice work @seshapad! We should also possibly add default for config file too?

done!

cmd/transform.go
seshapad: done!

cmd/transform.go
seshapad: done!

common/pathsofinterest.go
seshapad: done!

ashokponkumar (approved):

PR #809, opened 2022-05-17
fix: check properly if the default files exist and ignore if they don't

HarikrishnanBalagopal:

### DIFF
--- a/common/pathsofinterest.go
+++ b/common/pathsofinterest.go
@@ -28,7 +28,7 @@ const (
 	// DefaultConfigFilePath defines default config file path
 	DefaultConfigFilePath = types.AppNameShort + "-default-config.yaml"
 	// DefaultCustomizationFolder defines default name for plan file
-	DefaultCustomizationFolder = types.AppNameShort + "default-customizations"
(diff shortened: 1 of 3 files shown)


common/pathsofinterest.go
```diff
@@ -28,7 +28,7 @@ const (
 	// DefaultConfigFilePath defines default config file path
 	DefaultConfigFilePath = types.AppNameShort + "-default-config.yaml"
 	// DefaultCustomizationFolder defines default name for plan file
-	DefaultCustomizationFolder = types.AppNameShort + "-customization"
+	DefaultCustomizationFolder = types.AppNameShort + "default-customizations"
```
ashokponkumar, line 31: do we want a hyphen before the default?

ashokponkumar (changes requested):
Looks good. Small nit. @seshapad PTAL

common/pathsofinterest.go
HarikrishnanBalagopal: done

seshapad (approved):

ashokponkumar (approved):

PR #815, opened 2022-06-08
feat: User-defined label for cluster related QA

seshapad:


common/constants.go
```diff
@@ -130,7 +130,7 @@ const (
 	//ConfigIngressKey represents Ingress Key
 	ConfigIngressKey = ConfigTargetKey + d + "ingress"
 	//ConfigIngressHostKey represents Ingress host Key
-	ConfigIngressHostKey = ConfigIngressKey + d + "host"
+	ConfigIngressHostKey = "ingress" + d + "host"
```
ashokponkumar, line 133: Why do we want to hardcode the key here?

types/collection/cluster.go
```diff
@@ -25,6 +25,9 @@ import (
 // ClusterMetadataKind defines the kind of cluster metadata file
 const ClusterMetadataKind types.Kind = "ClusterMetadata"
 
+// QaLabelKey is the keyname for the qaLabel
+const QaLabelKey = "qalabel"
```
ashokponkumar, line 29: we might want to use a more elaborate label like move2kube.konveyor.io/clusterqalabel

ashokponkumar (changes requested):

ashokponkumar (approved):

PR #817, opened 2022-06-21
fix: binaries compressed for Ubuntu seg fault when run on CentOS

HarikrishnanBalagopal:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -165,7 +165,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: ${{ steps.info.outputs.go_version }}
-      - run: VERSION='${{ github.event.inputs.tag }}' HAS_UPX='false' make dist
+      - run: VERSION='${{ github.event.inputs.tag }}' make dist
       - name: upload release assets
         uses: konveyor/upload-release-action@v3
         with:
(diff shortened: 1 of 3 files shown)


ashokponkumar:
Was thinking we can remove the entire upx stuff from our code, including Makefiles. Anyway, we use zip for our releases, so it should compress reducing the download. 

Thoughts?

HarikrishnanBalagopal:
> Was thinking we can remove the entire upx stuff from our code, including Makefiles. Anyway, we use zip for our releases, so it should compress reducing the download.
> 
> Thoughts?

It doesn't hurt to have it. Still works when you are compiling locally.

ashokponkumar:
The main benefit was network download speed, which is anyway optimised by the archive format. 

If upx is present locally, our makefile will start using it. We won't know the version of upx and it might cause errors and there is no real benefit to them if built locally, since a 40 MB addition is not big for a hard disk. Also, I would assume compressing will increase execution time at least by a bit.

HarikrishnanBalagopal:
> If upx is present locally, our makefile will start using it.

Yeah and that's fine because it works locally

HarikrishnanBalagopal:
> The main benefit was network download speed, which is anyway optimised by the archive format.
> 
> If upx is present locally, our makefile will start using it. We won't know the version of upx and it might cause errors and there is no real benefit to them if built locally, since a 40 MB addition is not big for a hard disk. Also, I would assume compressing will increase execution time at least by a bit.

I have removed it entirely.

ashokponkumar (approved):

PR #820, opened 2022-06-30
feat: allow creating one dockerfile per child project in a dot net multi-project app

HarikrishnanBalagopal:

### DIFF
--- a/transformer/dockerfilegenerator/dotnet/constants.go
+++ b/transformer/dockerfilegenerator/dotnet/constants.go
@@ -1,0 +1,32 @@
+/*
+ *  Copyright IBM Corporation 2022
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package dotnet
+
+import (
+	transformertypes "github.com/konveyor/move2kube/types/transformer"
+)
+
+const (
+	// CSPROJ_FILE_EXT is the file extension for C# (C Sharp) projects.
+	CSPROJ_FILE_EXT = ".csproj"
+	// LaunchSettingsJSON is the name of the json containing launch configuration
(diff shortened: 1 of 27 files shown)


transformer/dockerfilegenerator/dotnet/utils.go
```diff
@@ -0,0 +1,58 @@
...
+
+type buildOption string
+
+const (
+	// NO_BUILD_STAGE don't generate the build stage in Dockerfiles
+	NO_BUILD_STAGE buildOption = "no-build-stage"
+	// BUILD_IN_BASE_IMAGE generate the build stage and put it in a separate Dockerfile
+	BUILD_IN_BASE_IMAGE buildOption = "build-stage-in-base-image"
+	// BUILD_IN_EVERY_IMAGE generate the build stage in every Dockerfile
+	BUILD_IN_EVERY_IMAGE buildOption = "build-stage-in-every-image"
```
ashokponkumar, line 34: Why not use normal non-hyphen strings?

ashokponkumar (changes requested):
Nice work! small suggestion.

transformer/dockerfilegenerator/dotnet/constants.go
```diff
@@ -0,0 +1,32 @@
...
+
+const (
+	// CSPROJ_FILE_EXT is the file extension for C# (C Sharp) projects.
+	CSPROJ_FILE_EXT = ".csproj"
+	// LaunchSettingsJSON is the name of the json containing launch configuration
+	LaunchSettingsJSON = "launchSettings.json"
+	// DotNetCoreCsprojFilesPathType points to the csproj files path of dotnetcore projects
+	DotNetCoreCsprojFilesPathType transformertypes.PathType = "DotNetCoreCsprojPathType"
+	// DotNetCoreSolutionFilePathType points to the solution file path of dotnetcore project
+	DotNetCoreSolutionFilePathType transformertypes.PathType = "DotNetCoreSolutionPathType"
```
Akash-Nayak, line 31: Nice work @HarikrishnanBalagopal! Just one small thing which is not clear to me that why are we placing these `dotnetcore` constants in the `dotnet` constants.go? Like, the `launchSettings.json` file doesn't exist for a dotnet project, I think it's only in dotnetcore.

transformer/dockerfilegenerator/dotnet/utils.go
HarikrishnanBalagopal: done

transformer/dockerfilegenerator/dotnet/constants.go
HarikrishnanBalagopal: I think we can merge all the different dot net transformers since they share so much logic.

ashokponkumar (approved):

PR #821, opened 2022-07-01
fix: the run stage of dockerfiles generated for dot net web app was using a deprecated base image

HarikrishnanBalagopal:

### DIFF
--- a/transformer/dockerfilegenerator/windows/webappdockerfilegenerator.go
+++ b/transformer/dockerfilegenerator/windows/webappdockerfilegenerator.go
@@ -39,21 +39,8 @@ const (
 	defaultRunStageImageTag = "4.8"
 )
 
-var (
-	// imageTagToSupportedVersions is a mapping from image tag to dot net framework versions that image supports.
-	// Version compatibility table taken from https://hub.docker.com/_/microsoft-dotnet-framework-aspnet
-	imageTagToSupportedVersions = map[string][]string{
-		"4.8":   {"4.8"},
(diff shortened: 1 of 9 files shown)


transformer/dockerfilegenerator/windows/webappdockerfilegenerator.go
```diff
@@ -34,19 +34,43 @@ import (
...
-	defaultBuildOutputDir = "bin"
+	buildStageC             = "dotnetwebbuildstage"
+	defaultBuildOutputDir   = "bin"
+	defaultRunStageImageTag = "4.8"
 )
 
+var (
+	// imageTagToSupportedVersions is a mapping from image tag to dot net framework versions that image supports.
+	// Version compatibility table taken from https://hub.docker.com/_/microsoft-dotnet-framework-aspnet
+	imageTagToSupportedVersions = map[string][]string{
```
ashokponkumar, line 45: Can we read this from a config file similar to how we do in java

ashokponkumar (changes requested):

transformer/dockerfilegenerator/windows/webappdockerfilegenerator.go
ashokponkumar: The advantage it provides is that it makes the mapping configurable without the need to change code.

HarikrishnanBalagopal: done

Issue #790, opened 2022-03-23
bug: ingress missing ingress class name


HarikrishnanBalagopal:
# Bug
## Overview
The Ingress yaml that `move2kube transform` generates is missing the `ingressClassName` field.
The officially supported ingress controller https://github.com/kubernetes/ingress-nginx adds an ingress class called `nginx` , but this is not set as the default ingress class.
Without an Ingress class, the deployed ingress doesn't get assigned an address and doesn't become available.

The `nginx` ingress class doesn't come with the `ingressclass.kubernetes.io/is-default-class: "true"` annotation.
https://kubernetes.io/docs/concepts/services-networking/ingress/#default-ingress-class
`$ kubectl get ingressclasses.networking.k8s.io nginx -o yaml`
```yaml
apiVersion: networking.k8s.io/v1
kind: IngressClass
metadata:
  annotations:
    meta.helm.sh/release-name: ingress-nginx
    meta.helm.sh/release-namespace: ingress-nginx
  creationTimestamp: "2022-03-23T06:16:35Z"
  generation: 1
  labels:
    app.kubernetes.io/component: controller
    app.kubernetes.io/instance: ingress-nginx
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: ingress-nginx
    app.kubernetes.io/version: 1.1.0
    helm.sh/chart: ingress-nginx-4.0.13
  name: nginx
  resourceVersion: "300650"
  uid: 4d7597e0-2ab8-4228-b310-5af35e16cdbe
spec:
  controller: k8s.io/ingress-nginx
```
## Steps To Reproduce
Run `move2kube transform` on a sample app like [language-platforms](https://github.com/konveyor/move2kube-demos/tree/main/samples/language-platforms)

```
$ move2kube plan -s language-platforms/
$ move2kube transform --qa-skip
```

## Actual Output
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  creationTimestamp: null
  labels:
    move2kube.konveyor.io/service: myproject
  name: myproject
spec:
  rules:
    - host: myproject.com
      http:
        paths:
          - backend:
              service:
                name: myproject-django
                port:
                  name: port-8080
            path: /myproject-django
            pathType: Prefix
          - backend:
              service:
                name: myproject-python
                port:
                  name: port-8080
            path: /myproject-python
            pathType: Prefix
          - backend:
              service:
                name: nodejs
                port:
                  name: port-8080
[...]

PR #822, opened 2022-07-04
fix: IngressClassName missing #790

bartvandongen3:
Add ingressClassName. Using the qaengine to set the IngressClassName and the default value is nginx. 

issue was: 
https://github.com/konveyor/move2kube/issues/790

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ get: go.mod
 	go mod download
 
 .PHONY: generate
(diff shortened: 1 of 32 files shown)


ashokponkumar (changes requested):
Thanks for the PR! Few suggestions.

ashokponkumar (approved):

PR #823, opened 2022-07-06
fix: panic when docker is not installed and CNB containerizer is used

HarikrishnanBalagopal:
New output after the fix
```
$ move2kube plan -s src/ --preset enablecontainerizedtransformers
INFO[0000] Pulling container image quay.io/konveyor/hello-world. This could take a few mins. 
FATA[0000] failed to initialize the container engine. Error: "failed to use docker as the container engine. Error: \"unable to run test image 'quay.io/konveyor/hello-world' as a container. Error: \\\"failed to pull the image 'quay.io/konveyor/hello-world'. Error: \\\\\\\"failed to pull the image 'quay.io/konveyor/hello-world' using the docker client. Error: \\\\\\\\\\\\\\\"Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?\\\\\\\\\\\\\\\"\\\\\\\"\\\"\""
```
We exit with error since no container engine is found.

### DIFF
--- a/environment/container/container.go
+++ b/environment/container/container.go
@@ -55,14 +55,11 @@ type ContainerEngine interface {
 func initContainerEngine() (err error) {
 	workingEngine, err = newDockerEngine()
 	if err != nil {
-		logrus.Debugf("Unable to use docker : %s", err)
-		return err
+		return fmt.Errorf("failed to use docker as the container engine. Error: %q", err)
 	}
 	//TODO: Add Support for podman
 	if workingEngine == nil {
-		err := fmt.Errorf("no working container runtime available")
-		logrus.Errorf("%s", err)
-		return err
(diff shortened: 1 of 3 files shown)


ashokponkumar (approved):

HarikrishnanBalagopal:
Relevant lines

We are not checking the error `initContainerEngine` returns
https://github.com/konveyor/move2kube/blob/main/environment/container/container.go#L75

`workingEngine` is nil because we return nil here https://github.com/konveyor/move2kube/blob/main/environment/container/dockerengine.go#L62

we are trying to return an error because `workingEngine` is nil,
but the bug is that `err` is also nil
https://github.com/konveyor/move2kube/blob/main/environment/container/container.go#L65

PR #824, opened 2022-07-07
Patch arch

kmehant:
WIP


HarikrishnanBalagopal (approved):
LGTM

Makefile
```diff
@@ -195,3 +196,25 @@ endif
 	# To help with reusing layers and hence speeding up build
 	${CONTAINER_TOOL} push ${REGISTRYNS}/${BINNAME}-builder:${VERSION}
 	${CONTAINER_TOOL} push ${REGISTRYNS}/${BINNAME}:${VERSION}
+
+.PHONY: cmultibuildpush
+cmultibuildpush: ## Build and push multi arch container image
+ifndef CONTAINER_TOOL
+	$(error No container tool (docker, podman) found in your environment. Please, install one)
```
ashokponkumar, line 203: can't we change the check to be for docker alone, then we can remove the checks below too.

ashokponkumar (changes requested):
Nice work!

PR #830, opened 2022-07-15
fix: workflow uses multi arch build and push, cache from latest, handle platform args

kmehant:


kmehant:
@ashokponkumar please review !!

ashokponkumar (approved):

Issue #811, opened 2022-05-23
The nodejs version is missing in the base image in the Move2Kube generated Dockerfile for the Frontend service of enterprise-app


Akash-Nayak:
**Describe the bug**
The frontend app's package.json file was updated to restrict the node version <v16.15. https://github.com/konveyor/move2kube-demos/blob/1c4d39c1bc887f44686ac0356795f92f3d1200e6/samples/enterprise-app/src/frontend/package.json#L10
 Now, in the Move2Kube generated Dockerfile for Frontend service of enterprise-app the node version in the base image is missing.
 
Also, the available Node.js versions in registry.access.redhat.com are 10, 12, 14 and 16 (10 & 12 are deprecated versions). So, Move2Kube should only take one of these available versions else, the image build may fail.

**To Reproduce**
1. Run `move2kube transform -s src --qa-skip` on the src folder of the enterprise-app.
2. Check the Dockerfile generated for the frontend. `FROM registry.access.redhat.com/ubi8/nodejs-` is missing the Node.Js version.

```console
$ cat myproject/source/frontend/Dockerfile
FROM registry.access.redhat.com/ubi8/nodejs-
COPY . .
RUN npm install
RUN npm run build
EXPOSE 8080
CMD npm run start
```

**Expected behavior**
The generated Dockerfile should have the suitable node version in it.

```console
$ cat myproject/source/frontend/Dockerfile
FROM registry.access.redhat.com/ubi8/nodejs-14
COPY . .
RUN npm install
RUN npm run build
EXPOSE 8080
CMD npm run start
```

**Desktop (please complete the following information):**
 - OS: Mac
 - Move2Kube Version: v0.3.3-rc.1

Akash-Nayak:
This PR https://github.com/konveyor/move2kube/pull/814 puts a temporary fix for the missing Node.js version in the generated Dockerfile which happens in case of conditional node version (say, <16.15) in the package.json file. This will prevent the image build from failing.

```console
FROM registry.access.redhat.com/ubi8/nodejs-14
COPY . .
RUN npm install
RUN npm run build
EXPOSE 8080
CMD npm run start
```

A better heuristic for handing the conditional Node.Js version in the package.json can also be added. So, this issue can be kept open till then.

Akash-Nayak:
https://github.com/konveyor/move2kube/pull/834 resolved this issue.

PR #834, opened 2022-07-18
feat: Add node version selection heuristic to handle conditional Node.JS version in package.json

Akash-Nayak:
Resolves- https://github.com/konveyor/move2kube/issues/811

Co-authored-by: Harikrishnan Balagopal <harikrishmenon@gmail.com>

### DIFF
--- a/transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
+++ b/transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
@@ -19,35 +19,44 @@ package dockerfilegenerator
 import (
 	"fmt"
 	"path/filepath"
-	"sort"
-	"strconv"
-	"strings"
 
-	"github.com/hashicorp/go-version"
 	"github.com/joho/godotenv"
 	"github.com/konveyor/move2kube/common"
 	"github.com/konveyor/move2kube/environment"
+	"github.com/konveyor/move2kube/types"
 	irtypes "github.com/konveyor/move2kube/types/ir"
 	"github.com/konveyor/move2kube/types/qaengine/commonqa"
 	transformertypes "github.com/konveyor/move2kube/types/transformer"
 	"github.com/konveyor/move2kube/types/transformer/artifacts"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cast"
-	"golang.org/x/mod/semver"
 )
 
 const (
-	defaultNodeVersion = "14"
-	packageJSONFile    = "package.json"
+	packageJSONFile        = "package.json"
+	versionMappingFilePath = "mappings/nodeversions.yaml"
+	// NodeVersionsMappingKind defines kind of NodeVersionMappingKind
+	NodeVersionsMappingKind types.Kind = "NodeVersionsMapping"
 )
 
-var supportedNodeVersions = []int{14, 16, 12, 10}
+// NodeVersionsMapping stores the Node versions mapping
+type NodeVersionsMapping struct {
+	types.TypeMeta   `yaml:",inline"`
+	types.ObjectMeta `yaml:"metadata,omitempty"`
+	Spec             NodeVersionsMappingSpec `yaml:"spec,omitempty"`
+}
+
+// NodeVersionsMappingSpec stores the Node version spec
+type NodeVersionsMappingSpec struct {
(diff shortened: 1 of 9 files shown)


transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
```diff
@@ -118,6 +163,54 @@ func (t *NodejsDockerfileGenerator) DirectoryDetect(dir string) (services map[st
 	return services, nil
 }
 
+// getNodeVersion returns the Node version to be used for the service
+func getNodeVersion(node string, allowedNodeVersions []int) string {
```
HarikrishnanBalagopal, line 167: Use https://github.com/hashicorp/go-version instead

HarikrishnanBalagopal (changes requested):

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
Akash-Nayak: The problem with using a library for parsing in this case is that, suppose if the version in package.json is `<16.15`, then I think that the library-based parser would return the version as 15. But, `registry.redhat.io/rhel8/nodejs-15` image does not exist in the Red Hat certified container images and the versions which are available are 10, 12, 14 and 16.

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
```diff
@@ -97,6 +98,50 @@ func (t *NodejsDockerfileGenerator) GetConfig() (transformertypes.Transformer, *
 	return t.Config, t.Env
 }
 
+// fundNearest returns the closest number in the arr to the given num
+func findNearest(arr []int, num int) int {
```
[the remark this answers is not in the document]
Akash-Nayak: Yes, the array is a small array, just as it was a sorted array so I used binary search instead of linear search.

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
HarikrishnanBalagopal: Example: https://go.dev/play/p/vdKWgIP0IwB

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
```diff
@@ -39,6 +41,8 @@ const (
 	packageJSONFile    = "package.json"
 )
 
+var supportedNodeVersions = []int{14, 16, 12, 10}
```
ashokponkumar, line 44: We should read this from a mapping file containing the version numbers and tag names, similar to what we do with java and dotnet.

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
```diff
@@ -97,6 +101,46 @@ func (t *NodejsDockerfileGenerator) GetConfig() (transformertypes.Transformer, *
...
+			nodeVersion = nodeVersion - 1
+		}
+		sort.Ints(supportedVersions)
+		if nodeVersion < supportedVersions[0] {
+			return strconv.Itoa(supportedVersions[0])
+		}
+		if nodeVersion > supportedVersions[len(supportedVersions)-1] {
+			return strconv.Itoa((supportedVersions[len(supportedVersions)-1]))
+		}
+		return strconv.Itoa(nodeVersion)
```
ashokponkumar, line 126: Instead of relying on specific idea like even version, we can possibly find the nearest (greater?) version in the list.

ashokponkumar (changes requested):
Few more suggestions..

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
Akash-Nayak: Done.

transformer/dockerfilegenerator/nodejsdockerfiletransformer.go
Akash-Nayak: Now, we go over the versions in the list, and which ever version first satisfies the constraint we select it.

HarikrishnanBalagopal (approved):
LGTM. Thanks!

PR #838, opened 2022-07-26
fix: ir was not being propagated through artifacts for use by downstream transformers

HarikrishnanBalagopal:


seshapad (approved):

ashokponkumar (changes requested):
We should do for war and ear too

ashokponkumar (approved):

PR #840, opened 2022-07-29
fix: detect the node version from the package.json in the dot net core transformer instead of hard coding

HarikrishnanBalagopal:

### DIFF
--- a/assets/built-in/transformers/dockerfilegenerator/mappings/nodeversions.yaml
+++ b/assets/built-in/transformers/dockerfilegenerator/mappings/nodeversions.yaml
@@ -4,7 +4,11 @@ metadata:
   name: NodeVersionsMapping
 spec:
   nodeVersions:
-    - "v16.16.0"
-    - "v14.20.0"
-    - "v12.22.12"
-    - "v10.24.1"
+    - version: "v16.16.0"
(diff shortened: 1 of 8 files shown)


assets/built-in/transformers/dockerfilegenerator/mappings/nodeversions.yaml
```diff
@@ -4,7 +4,11 @@ metadata:
   name: NodeVersionsMapping
 spec:
   nodeVersions:
-    - "v16.16.0"
-    - "v14.20.0"
-    - "v12.22.12"
-    - "v10.24.1"
+    "v16.16.0":
```
ashokponkumar, line 7: Putting it as a dictionary voids the order. That's why we might want to use a list.

ashokponkumar (changes requested):
Small suggestion..

assets/built-in/transformers/dockerfilegenerator/mappings/nodeversions.yaml
HarikrishnanBalagopal: we are sorting it so it doesn't matter

ashokponkumar (approved):

PR #844, opened 2022-08-04
fix: bug where running on Windows would produce paths with back slashes even though the Dockerfile uses a Linux base image

HarikrishnanBalagopal:


transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
```diff
@@ -438,9 +438,9 @@ func (t *DotNetCoreDockerfileGenerator) TransformArtifact(newArtifact transforme
 			buildOutputDir = filepath.Clean(common.GetUnixPath(configuration.PropertyGroups[idx].OutputPath))
 		}
 		relBuildOutputDir := filepath.Join(buildOutputDir, targetFramework, defaultBuildOutputSubDir)
```
ashokponkumar, line 440: Dont' we have to do it here too?

ashokponkumar (changes requested):

transformer/dockerfilegenerator/dotnetcoredockerfilegenerator.go
HarikrishnanBalagopal: No, because in that case we are copying from the local filesystem which should use the OS specific path separator

HarikrishnanBalagopal: have changed to unix paths

ashokponkumar (approved):

PR #855, opened 2022-08-26
fix: Update docker/distrbution dep to 2.8.1

jmontleon:
I can not build with docker/distribution 2.8.0. It looks like there was a problem that was corrected in 2.8.1
https://github.com/distribution/distribution/issues/3590

`verifying github.com/docker/distribution@v2.8.0+incompatible: checksum mismatch`


ashokponkumar:
Thanks. Can you prefix the pr name with "fix:"?

ashokponkumar (approved):

PR #859, opened 2022-09-05
fix: detect the java and app version for generating proper dockerfiles

HarikrishnanBalagopal:


ashokponkumar (commented):
Is there any other property similar like version in the properties file that we should be using?

ashokponkumar (approved):

PR #862, opened 2022-09-07
feat: support default transformers with no source input

kmehant:


cmd/transform.go
```diff
@@ -162,12 +161,16 @@ func transformHandler(cmd *cobra.Command, flags transformFlags) {
...
+		if p.Spec.SourceDir != "" {
+			checkSourcePath(p.Spec.SourceDir)
+		}
 		lib.CheckAndCopyCustomizations(p.Spec.CustomizationsDir)
 		flags.outpath = filepath.Join(flags.outpath, p.Name)
 		checkOutputPath(flags.outpath, flags.overwrite)
-		if p.Spec.SourceDir == flags.outpath || common.IsParent(flags.outpath, p.Spec.SourceDir) || common.IsParent(p.Spec.SourceDir, flags.outpath) {
-			logrus.Fatalf("The source path %s and output path %s overlap.", p.Spec.SourceDir, flags.outpath)
+		if p.Spec.SourceDir != "" {
+			if p.Spec.SourceDir == flags.outpath || common.IsParent(flags.outpath, p.Spec.SourceDir) || common.IsParent(p.Spec.SourceDir, flags.outpath) {
```
ashokponkumar, line 171: Can't we make it a single if command?

types/plan/plan.go
```diff
@@ -38,7 +38,8 @@ type Spec struct {
 	SourceDir         string `yaml:"sourceDir"`
 	CustomizationsDir string `yaml:"customizationsDir,omitempty"`
 
-	Services map[string][]PlanArtifact `yaml:"services"` //[servicename]
+	Services              map[string][]PlanArtifact `yaml:"services"`                                         //[servicename]
+	TransformersByDefault map[string]string         `yaml:"transformersByDefault,omitempty" m2kpath:"normal"` //[name]filepath
```
ashokponkumar, line 42: TransformersByDefault seems non-intuitive. How about something like `InvokedByDefaultTransformers`, `AlwaysRunTransformers` or `TransformersWithNoInput`? Also, since the transformer paths already come in the transformers section, here we can just list the transformer names alone.

types/plan/planutils.go
```diff
@@ -30,14 +30,14 @@ import (
...
 	var err error
+	absSourceDir := ""
 	if err = common.ReadMove2KubeYaml(path, &plan); err != nil {
 		logrus.Errorf("Failed to load the plan file at path %q Error %q", path, err)
 		return plan, err
 	}
-	if sourceDir != "" {
-		plan.Spec.SourceDir = sourceDir
+	if plan.Spec.SourceDir != "" {
+		absSourceDir, err = filepath.Abs(plan.Spec.SourceDir)
```
ashokponkumar, line 39: Why are we ignoring the use of sourceDir here?

types/transformer/transformer.go
```diff
@@ -46,6 +46,12 @@ type TransformerSpec struct {
 	OverrideSelector   labels.Selector                        `yaml:"-" json:"-"`
 	TemplatesDir       string                                 `yaml:"templates" json:"templates"` // Relative to yaml directory or working directory in image
 	Config             interface{}                            `yaml:"config" json:"config"`
+	InvokesByDefault   InvokesByDefault                       `yaml:"invokesByDefault" json:"invokesByDefault"`
+}
+
+// InvokesByDefault stores config to toggle default transformers
+type InvokesByDefault struct {
```
ashokponkumar, line 53: And Keep this name consistent with the name we use in the plan.

ashokponkumar (changes requested):
Nice work @kmehant ! Added some comments.

types/transformer/transformer.go
kmehant: Shall we leave this the same? I mean we are using `invokesByDefault` in the plan file which goes well with `InvokedByDefaultTransformers` too

kmehant: changed it to `InvokedByDefault`

types/plan/planutils.go
kmehant: Updated it1

types/plan/plan.go
kmehant: Thanks for letting me know! 

updated with `InvokedByDefaultTransformers` as it is more relevant as it provides the difference between transformers that are invoked by default or invoked due to services (through directory detect).

modified the section to just show the transformer names

cmd/transform.go
kmehant: Done1

PR #867, opened 2022-09-19
feat: add resources dir starlark custom variable

kmehant:


ashokponkumar (changes requested):
We should not be exposing the assets directory. Its internal to move2kube. Rather each transformer can use the context dir. context_dir

PR #870, opened 2022-09-20
feat: add archive utils with compression options

kmehant:

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -224,8 +224,8 @@ var PlanProgressNumDirectories = 0
 type CompressionType = string
 
 const (
-	// GZipCompressionType allows archival using gzip compression format
-	GZipCompressionType CompressionType = "GZip"
+	// GZipCompression allows archival using gzip compression format
+	GZipCompression CompressionType = "GZip"
 	// NoCompression allows archival without compression
 	NoCompression CompressionType = "None"
 )
--- a/common/constants.go
+++ b/common/constants.go
@@ -226,6 +226,6 @@ type CompressionType = string
 const (
 	// GZipCompressionType allows archival using gzip compression format
 	GZipCompressionType CompressionType = "GZip"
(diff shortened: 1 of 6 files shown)


ashokponkumar (changes requested):
Can we reuse https://github.com/konveyor/move2kube/blob/6784a98bc22efe016805975add636e0b5250e150/environment/container/utils.go#L106 function?

filesystem/templatecopy.go
```diff
@@ -169,6 +169,7 @@ func writeTemplateToFile(tpl string, config interface{}, writepath string,
 		"execTemplate":   execTemplate(packageTemplate),
 		"encAesCbcPbkdf": common.EncryptAesCbcWithPbkdfWrapper,
 		"encRsaCert":     common.EncryptRsaCertWrapper,
+		"archTarGZipStr": common.CreateTarArchiveGZipStringWrapper,
```
ashokponkumar, line 172: Would be good to expose just tar too.

common/constants.go
```diff
@@ -219,3 +219,13 @@ var PlanProgressNumBaseDetectTransformers = 0
...
 var PlanProgressNumDirectories = 0
+
+// CompressionType refers to the compression type
+type CompressionType = string
+
+const (
+	// GZipCompressionType allows archival using gzip compression format
+	GZipCompressionType CompressionType = "GZip"
+	// DefaultCompressionType allows archival without compression
+	DefaultCompressionType CompressionType = "Default"
```
ashokponkumar, line 230: the default is no compression, isn't it? I assume Tar is not a compression format. So can we possibly call it as `NoCompression` = `None`?

ashokponkumar (changes requested):
Nice work. Some nit.

ashokponkumar (approved):

PR #871, opened 2022-09-21
feat: add starlark custom functions for crypto and archive utils

kmehant:

### DIFF
--- a/transformer/external/starlarktransformer.go
+++ b/transformer/external/starlarktransformer.go
@@ -70,11 +70,11 @@ const (
 	fspathRelFnName              = "pathrel"
 
 	// encryption functions
-	encaescbcpbkdfFnName = "encaescbcpbkdf"
-	encrsacertFnName     = "encrsacert"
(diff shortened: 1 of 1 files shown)


transformer/external/starlarktransformer.go
```diff
@@ -68,6 +68,13 @@ const (
 	fswriteFnName                = "write"
 	fspathBaseFnName             = "pathbase"
 	fspathRelFnName              = "pathrel"
+
+	// encryption functions
+	encaescbcpbkdfFnName = "encaescbcpbkdf"
```
ashokponkumar, line 73: I think we have to reevaluate starlark function naming conventions. These function names are almost unreadable. Can you please check the conventions and come with with alternatives

ashokponkumar (changes requested):

transformer/external/starlarktransformer.go
kmehant: Shall we track this in a separate PR?

ashokponkumar (approved):

PR #876, opened 2022-09-27
fix: fail when none of the source or customizations folder is given

kmehant:

### DIFF
--- a/cmd/plan.go
+++ b/cmd/plan.go
@@ -140,6 +140,10 @@ func planHandler(cmd *cobra.Command, flags planFlags) {
 		return
 	}
 	logrus.Infof("Plan can be found at [%s].", planfile)
+	if len(p.Spec.Services) == 0 && len(p.Spec.InvokedByDefaultTransformers) == 0 {
+		logrus.Debugf("Plan : %+v", p)
+		logrus.Fatalf("failed to find any services or default transformers.")
+	}
 }
 
 // GetPlanCommand returns a command to do the planning
--- a/cmd/transform.go
+++ b/cmd/transform.go
@@ -136,6 +136,10 @@ func transformHandler(cmd *cobra.Command, flags transformFlags) {
 		startQA(flags.qaflags)
 		logrus.Debugf("Creating a new plan.")
 		p = lib.CreatePlan(ctx, flags.srcpath, flags.outpath, flags.customizationsPath, flags.transformerSelector, flags.name)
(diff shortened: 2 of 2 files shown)


lib/planner.go
```diff
@@ -28,6 +28,9 @@ import (
 
 //CreatePlan creates the plan from all planners
 func CreatePlan(ctx context.Context, inputPath, outputPath string, customizationsPath, transformerSelector, prjName string) plantypes.Plan {
+	if inputPath == "" && customizationsPath == "" {
```
ashokponkumar, line 31: Why should it fail when both of these are empty? Even internal transformers can be invoked by default, isn't it

ashokponkumar (changes requested):

lib/planner.go
kmehant: Ah, you are right!

lib/planner.go
```diff
@@ -60,6 +60,10 @@ func CreatePlan(ctx context.Context, inputPath, outputPath string, customization
 	}
 	logrus.Infoln("Configuration loading done")
 
+	if len(p.Spec.Services) == 0 && len(p.Spec.InvokedByDefaultTransformers) == 0 {
```
ashokponkumar, line 63: Won't the list of services only become available only after the planning is done below?

ashokponkumar (changes requested):

lib/planner.go
kmehant: fixed it 😄

HarikrishnanBalagopal (approved):
LGTM

cmd/plan.go
```diff
@@ -140,6 +140,10 @@ func planHandler(cmd *cobra.Command, flags planFlags) {
 		return
 	}
 	logrus.Infof("Plan can be found at [%s].", planfile)
+	if len(p.Spec.Services) == 0 && len(p.Spec.InvokedByDefaultTransformers) == 0 {
+		logrus.Debugf("Plan : %+v", p)
+		logrus.Fatalf("Failed to find any services or default transformers. Aborting.")
```
ashokponkumar, line 145: Do we need the Aborting word? Since it has actually finished the process.

ashokponkumar (changes requested):

cmd/plan.go
kmehant: correct, as we are creating the plan file. Fixed it.

ashokponkumar (approved):

PR #877, opened 2022-09-27
fix: graph support for default transformers

kmehant:


kmehant:
Screenshot 2022-09-27 at 9 21 05 PM

transformer/transformer.go
```diff
@@ -558,20 +563,20 @@ func transform(newArtifactsToProcess, allArtifacts []transformertypes.Artifact,
 	return pathMappings, newArtifactsCreated, nil
 }
 
-func runSingleTransform(artifactsToProcess, allArtifacts []transformertypes.Artifact, transformer Transformer, tconfig transformertypes.Transformer, env *environment.Environment, graph *graphtypes.Graph, iteration int) ([]transformertypes.PathMapping, []transformertypes.Artifact, error) {
+func runSingleTransform(artifactsToProcess, allArtifacts []transformertypes.Artifact, transformer Transformer, tconfig transformertypes.Transformer, env *environment.Environment, graph *graphtypes.Graph, iteration int) ([]transformertypes.PathMapping, []transformertypes.Artifact, int, error) {
```
ashokponkumar, line 566: Would be good to use named returns in the function definition to make it readable on what the 3rd int value represents.

ashokponkumar (changes requested):

transformer/transformer.go
kmehant: Sure Ashok, thats vertex id, should I define the type vertexid int and change everywhere its being used?

ashokponkumar: No. https://www.geeksforgeeks.org/named-return-parameters-in-golang/amp/

ashokponkumar (approved):

HarikrishnanBalagopal:
> Screenshot 2022-09-27 at 9 21 05 PM

is this the bug or final output after the PR?
Shouldn't the start be connected to the transformer?

kmehant:
New Graph:

Screenshot 2022-09-28 at 1 47 57 PM

transformer/transformer.go
```diff
@@ -580,6 +581,11 @@ func runSingleTransform(artifactsToProcess, allArtifacts []transformertypes.Arti
 				"pathMappings":      summarizePathMappings(newPathMappings),
 			},
 		)
+		// transformers that are invoked by default has source vertex as start
+		if tconfig.Spec.InvokedByDefault.Enabled {
+			edgeName := fmt.Sprintf("%d -> %d (invoked by default)", 0, targetVertexId)
+			graph.AddEdge(0, targetVertexId, edgeName, nil)
```
HarikrishnanBalagopal, line 587: Instead of hard coding `0`  as the start vertex Id, we can make it a constant that is shared between all the packages that need it.

HarikrishnanBalagopal (changes requested):
Just a small change

HarikrishnanBalagopal (approved):
LGTM

PR #878, opened 2022-09-27
feat: let the user decide if they want planning to fail if no services are detected

HarikrishnanBalagopal:

### DIFF
--- a/cmd/plan.go
+++ b/cmd/plan.go
@@ -144,9 +144,9 @@ func planHandler(cmd *cobra.Command, flags planFlags) {
 	logrus.Infof("Plan can be found at [%s].", planfile)
 	if len(p.Spec.Services) == 0 && len(p.Spec.InvokedByDefaultTransformers) == 0 {
 		if flags.failOnEmptyPlan {
-			logrus.Fatalf("Did not detect any services in the directory %s", srcpath)
+			logrus.Fatalf("Did not detect any services in the directory %s . Also we didn't find any default transformers to run.", srcpath)
 		}
-		logrus.Warnf("Did not detect any services in the directory %s", srcpath)
+		logrus.Warnf("Did not detect any services in the directory %s . Also we didn't find any default transformers to run.", srcpath)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


anwarchk:
If I understand it correctly, `move2kube` doesn't do anything with the bound services present in the cf manifest. So is this change from a transformers perspective ?

ashokponkumar:
> If I understand it correctly, `move2kube` doesn't do anything with the bound services present in the cf manifest. So is this change from a transformers perspective ?

@anwarchk This particular PR has nothing to do with Cf bound services. But we are currently working on a capability to migrate bound services to either as Operators or services in Cloud catalogs. [Here](https://kubernetes.slack.com/archives/CR85S82A2/p1663827038349669) is a discussion on this topic. 

Please do let us know your usecase, how you are using Move2Kube, and how you would like Move2Kube transform the services, so that we can feed that into feature design for services migration. We can also continue the conversation in [slack](https://kubernetes.slack.com/archives/CR85S82A2).

cmd/plan.go
```diff
@@ -139,10 +140,13 @@ func planHandler(cmd *cobra.Command, flags planFlags) {
 		logrus.Errorf("Unable to write plan file (%s) : %s", planfile, err)
 		return
 	}
+	logrus.Debugf("Plan : %+v", p)
 	logrus.Infof("Plan can be found at [%s].", planfile)
 	if len(p.Spec.Services) == 0 && len(p.Spec.InvokedByDefaultTransformers) == 0 {
-		logrus.Debugf("Plan : %+v", p)
-		logrus.Fatalf("failed to find any services or default transformers.")
+		if flags.failOnEmptyPlan {
+			logrus.Fatalf("Did not detect any services in the directory %s", srcpath)
```
ashokponkumar, line 147: we might also want to mention about default transformers similar to the previous message.

HarikrishnanBalagopal: Done

ashokponkumar (changes requested):

ashokponkumar (approved):

PR #883, opened 2022-09-29
chore: update the go.mod file to fix minor issues

HarikrishnanBalagopal:

### DIFF
--- a/transformer/transformer.go
+++ b/transformer/transformer.go
@@ -372,12 +372,14 @@ func walkForServices(inputPath string, bservices map[string][]plantypes.PlanArti
 			logrus.Debugf("[%s] Done", config.Name)
 			numfound += len(newPlanServices)
 			if len(newPlanServices) > 0 {
+				msg := getNamedAndUnNamedServicesLogMessage(newPlanServices)
 				relpath, err := filepath.Rel(inputPath, path)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


transformer/transformer.go
```diff
@@ -370,9 +370,13 @@ func walkForServices(inputPath string, bservices map[string][]plantypes.PlanArti
...
 			logrus.Debugf("[%s] Done", config.Name)
-			numfound += len(newServicesToArtifacts)
-			if len(newServicesToArtifacts) > 0 {
-				relpath, _ := filepath.Rel(inputPath, path)
+			numfound += len(newPlanServices)
+			if len(newPlanServices) > 0 {
+				relpath, err := filepath.Rel(inputPath, path)
+				if err != nil {
+					logrus.Errorf("failed to make the directory %s relative to the input directory %s . Error: %q", path, inputPath, err)
+					continue
```
ashokponkumar, line 378: Since it is only for log, we need not continue, but rather print error and proceed to log.

ashokponkumar (changes requested):

transformer/transformer.go
HarikrishnanBalagopal: Even if it is only for log it should not fail.

ashokponkumar (approved):

PR #888, opened 2022-10-07
fix: deselect the console and silverlight transformer by default to avoid conflicts with dotnet webapp transformer

HarikrishnanBalagopal:
Signed-off-by: Harikrishnan Balagopal <harikrishmenon@gmail.com>

PR #889, opened 2022-10-07
fix: deselect the console and silverlight transformer by default to avoid conflicts with dotnet webapp transformer

HarikrishnanBalagopal:
See https://github.com/konveyor/move2kube/pull/888

### DIFF
--- a/transformer/transformer.go
+++ b/transformer/transformer.go
@@ -37,12 +37,14 @@ import (
 	"github.com/konveyor/move2kube/transformer/dockerfilegenerator/windows"
 	"github.com/konveyor/move2kube/transformer/external"
 	"github.com/konveyor/move2kube/transformer/kubernetes"
+	"github.com/konveyor/move2kube/types"
 	environmenttypes "github.com/konveyor/move2kube/types/environment"
 	graphtypes "github.com/konveyor/move2kube/types/graph"
 	plantypes "github.com/konveyor/move2kube/types/plan"
 	transformertypes "github.com/konveyor/move2kube/types/transformer"
 	"github.com/konveyor/move2kube/types/transformer/artifacts"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/cast"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 )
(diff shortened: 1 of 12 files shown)


HarikrishnanBalagopal:
>  Yes, but it would be good to provide a way to list all transformers

Done

types/plan/plan.go
```diff
@@ -38,11 +38,12 @@ type Spec struct {
...
-	Services                     map[string][]PlanArtifact `yaml:"services"` //[servicename]
-	InvokedByDefaultTransformers []string                  `yaml:"invokedByDefaultTransformers,omitempty"`
+	Services map[string][]PlanArtifact `yaml:"services"` //[servicename]
 
-	TransformerSelector metav1.LabelSelector `yaml:"transformerSelector,omitempty"`
-	Transformers        map[string]string    `yaml:"transformers,omitempty" m2kpath:"normal"` //[name]filepath
+	TransformerSelector          metav1.LabelSelector `yaml:"transformerSelector,omitempty"`
+	Transformers                 map[string]string    `yaml:"transformers,omitempty" m2kpath:"normal"` //[name]filepath
+	InvokedByDefaultTransformers []string             `yaml:"invokedByDefaultTransformers,omitempty"`
+	DisabledTransformers         []string             `yaml:"disabledTransformers,omitempty"`
```
ashokponkumar, line 46: It might be good to have the path here too. Probably as a 2d array to allow for more than one path for a transformer.

transformer/transformer.go
```diff
@@ -176,11 +184,27 @@ func InitTransformers(transformerToInit map[string]string, selector labels.Selec
...
 	sort.Strings(transformerNames)
-	selectedTransformerNames := qaengine.FetchMultiSelectAnswer(common.ConfigTransformerTypesKey, "Select all transformer types that you are interested in:", []string{"Services that don't support any of the transformer types you are interested in will be ignored."}, transformerNames, transformerNames)
+	selectedTransformerNames := qaengine.FetchMultiSelectAnswer(
+		common.ConfigTransformerTypesKey,
+		"Select all transformer types that you are interested in:",
+		[]string{"Services that don't support any of the transformer types you are interested in will be ignored."},
+		defaultSelectedTransformerNames,
+		transformerNames,
+	)
+	deselectedTransformerNames := []string{}
```
ashokponkumar, line 202: If we are adding a section for disabled transformers, it might be good to add those disabled because of filters and errors too.

ashokponkumar (changes requested):

transformer/transformer.go
HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #896, opened 2022-10-12
feat: support qa engine answer validation and retries

kmehant:

### DIFF
--- a/qaengine/cliengine.go
+++ b/qaengine/cliengine.go
@@ -81,8 +81,14 @@ func (*CliEngine) fetchSelectAnswer(prob qatypes.Problem) (qatypes.Problem, erro
 		Options: prob.Options,
 		Default: def,
 	}
-	if err := survey.AskOne(prompt, &ans); err != nil {
-		logrus.Fatalf("Error while asking a question : %s", err)
+	if prob.Validation != nil {
+		if err := survey.AskOne(prompt, &ans, survey.WithValidator(prob.Validation)); err != nil {
+			logrus.Fatalf("Error while asking a question : %s", err)
+		}
+	} else {
+		if err := survey.AskOne(prompt, &ans); err != nil {
+			logrus.Fatalf("Error while asking a question : %s", err)
+		}
 	}
 	prob.Answer = ans
 	return prob, nil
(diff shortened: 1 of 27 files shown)


types/qaengine/qagrpc/fetchanswer.pb.go
```diff
@@ -64,6 +64,7 @@ type Problem struct {
 	Hints       []string `protobuf:"bytes,4,rep,name=hints,proto3" json:"hints,omitempty"`
 	Options     []string `protobuf:"bytes,5,rep,name=options,proto3" json:"options,omitempty"`
 	Default     []string `protobuf:"bytes,6,rep,name=default,proto3" json:"default,omitempty"`
+	Pattern string `protobuf:"bytes,7,rep,name=pattern,proto3" json:"pattern,omitempty"`
```
ashokponkumar, line 67: We should not be editing this directly. Only the proto file should be edited and this should be generated.

types/qaengine/problem.go
```diff
@@ -69,14 +71,28 @@ func NewProblem(p *qagrpc.Problem) (prob Problem, err error) {
...
 		Default: defaults,
-	}, nil
+	}
+	if p.Pattern != "" {
+		pp.Validation = func(ans interface{}) error {
+			reg, err := regexp.Compile(p.Pattern)
+			if err != nil {
+				return err
+			}
+			a := ans.(string)
```
ashokponkumar, line 88: if the answer type is not string, which is the case with question of type like multiselect, etc, this will panic, isn't it?

Also, don't we want validation for other types of questions too?

qaengine/cliengine.go
```diff
@@ -81,8 +81,14 @@ func (*CliEngine) fetchSelectAnswer(prob qatypes.Problem) (qatypes.Problem, erro
 		Options: prob.Options,
 		Default: def,
 	}
-	if err := survey.AskOne(prompt, &ans); err != nil {
-		logrus.Fatalf("Error while asking a question : %s", err)
+	if prob.Validation != nil {
```
ashokponkumar, line 84: Dont' we want the validation in httprestengine, defaultengine and storeengines too?

ashokponkumar (changes requested):
Nice work. Few comments.

PR #907, opened 2022-11-03
feat: Added containerized tranformer input/output configurability

seshapad:

### DIFF
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -79,8 +79,7 @@ type EnvironmentInstance interface {
 	Stat(name string) (fs.FileInfo, error)
 	Download(envpath string) (outpath string, err error)
 	Upload(outpath string) (envpath string, err error)
-	AddEnvironmentVariablesToInstance(envList []string) error
-	Exec(cmd environmenttypes.Command) (stdout string, stderr string, exitcode int, err error)
+	Exec(cmd environmenttypes.Command, envList []string) (stdout string, stderr string, exitcode int, err error)
 	Destroy() error
 
 	GetSource() string
@@ -164,11 +163,11 @@ func (e *Environment) Reset() error {
 }
 
 // Exec executes an executable within the environment
-func (e *Environment) Exec(cmd environmenttypes.Command) (stdout string, stderr string, exitcode int, err error) {
+func (e *Environment) Exec(cmd environmenttypes.Command, envList []string) (stdout string, stderr string, exitcode int, err error) {
 	if !e.active {
 		return "", "", 0, ErrEnvironmentNotActive
 	}
-	return e.Env.Exec(cmd)
+	return e.Env.Exec(cmd, envList)
 }
 
 // Destroy destroys all artifacts specific to the environment
--- a/environment/environment.go
+++ b/environment/environment.go
@@ -79,6 +79,7 @@ type EnvironmentInstance interface {
 	Stat(name string) (fs.FileInfo, error)
 	Download(envpath string) (outpath string, err error)
 	Upload(outpath string) (envpath string, err error)
+	AddEnvironmentVariablesToInstance(envList []string) error
 	Exec(cmd environmenttypes.Command) (stdout string, stderr string, exitcode int, err error)
 	Destroy() error
 
--- a/environment/local.go
+++ b/environment/local.go
@@ -86,14 +86,8 @@ func (e *Local) Stat(name string) (fs.FileInfo, error) {
 	return os.Stat(name)
 }
 
-// AddEnvironmentVariablesToInstance adds the environment variables after the environment is created
-func (e *Local) AddEnvironmentVariablesToInstance(envList []string) error {
-	e.EnvKeyValueList = append(e.EnvKeyValueList, envList...)
-	return nil
-}
-
 // Exec executes an executable within the environment
-func (e *Local) Exec(cmd environmenttypes.Command) (stdout string, stderr string, exitcode int, err error) {
+func (e *Local) Exec(cmd environmenttypes.Command, envList []string) (stdout string, stderr string, exitcode int, err error) {
 	if common.DisableLocalExecution {
 		return "", "", 0, fmt.Errorf("local execution prevented by %s flag", common.DisableLocalExecutionFlag)
 	}
@@ -108,7 +102,7 @@ func (e *Local) Exec(cmd environmenttypes.Command) (stdout string, stderr string
 	execcmd.Stdout = &outb
 	execcmd.Stderr = &errb
 	execcmd.Env = e.getEnv()
-	execcmd.Env = append(execcmd.Env, e.EnvKeyValueList...)
+	execcmd.Env = append(execcmd.Env, envList...)
 	if err := execcmd.Run(); err != nil {
 		var ee *exec.ExitError
 		var pe *os.PathError
--- a/environment/local.go
+++ b/environment/local.go
@@ -86,6 +86,12 @@ func (e *Local) Stat(name string) (fs.FileInfo, error) {
 	return os.Stat(name)
 }
 
+// AddEnvironmentVariablesToInstance adds the environment variables after the environment is created
+func (e *Local) AddEnvironmentVariablesToInstance(envList []string) error {
+	e.EnvKeyValueList = append(e.EnvKeyValueList, envList...)
+	return nil
+}
+
 // Exec executes an executable within the environment
 func (e *Local) Exec(cmd environmenttypes.Command) (stdout string, stderr string, exitcode int, err error) {
 	if common.DisableLocalExecution {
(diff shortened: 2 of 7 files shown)


transformer/external/executabletransformer.go
```diff
@@ -32,20 +32,48 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	envMode      = "env"
+	argMode      = "arg"
```
ashokponkumar, line 37: Might be good to add a type for different modes, and also have a default.

transformer/external/executabletransformer.go
```diff
@@ -32,20 +32,48 @@ import (
...
+
 // ExecutableYamlConfig is the format of executable yaml config
 type ExecutableYamlConfig struct {
 	EnableQA           bool                       `yaml:"enableQA"`
 	Platforms          []string                   `yaml:"platforms"`
 	DirectoryDetectCMD environmenttypes.Command   `yaml:"directoryDetectCMD"`
 	TransformCMD       environmenttypes.Command   `yaml:"transformCMD"`
 	Container          environmenttypes.Container `yaml:"container,omitempty"`
+	DetectScriptIO     InputOutput                `yaml:"detectScriptIO"`
+	TransformScriptIO  InputOutput                `yaml:"transformScriptIO"`
```
ashokponkumar, line 76: Do we need separate variables for this? Wouldn't keeping this with the CMD or the environment like `container` make it more coherent?

transformer/external/executabletransformer.go
```diff
@@ -138,7 +165,24 @@ func (t *Executable) Transform(newArtifacts []transformertypes.Artifact, already
...
+		t.Env.Env.AddEnvironmentVariablesToInstance([]string{t.ExecConfig.TransformScriptIO.Input.EnvName + envDelimiter + t.ExecConfig.TransformScriptIO.Input.Path})
+	}
+	if t.ExecConfig.TransformScriptIO.Output.Mode == argMode {
+		cmdToRun = append(cmdToRun, "output="+t.ExecConfig.TransformScriptIO.Output.Path)
+	} else if t.ExecConfig.TransformScriptIO.Output.Mode == envMode {
+		t.Env.EnvKeyValueList = append(t.Env.EnvKeyValueList,
+			t.ExecConfig.TransformScriptIO.Output.EnvName+envDelimiter+t.ExecConfig.TransformScriptIO.Output.Path)
+		t.Env.Env.AddEnvironmentVariablesToInstance([]string{t.ExecConfig.TransformScriptIO.Output.EnvName + envDelimiter + t.ExecConfig.TransformScriptIO.Output.Path})
+	}
+	stdout, stderr, exitcode, err := t.Env.Exec(cmdToRun)
```
ashokponkumar, line 185: since the processing for transform and detect commands are same, might be better to put the processing in a function, and call here.

ashokponkumar (changes requested):
Thanks Nabha. Few suggestions.

environment/environment.go
```diff
@@ -79,6 +79,7 @@ type EnvironmentInstance interface {
 	Stat(name string) (fs.FileInfo, error)
 	Download(envpath string) (outpath string, err error)
 	Upload(outpath string) (envpath string, err error)
+	AddEnvironmentVariablesToInstance(envList []string) error
```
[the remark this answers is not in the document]
seshapad: @ashokponkumar I had initially thought of using the Exec, but, it t.Env.Exec does not inherit the **environment.EnvironmentInstance**. EnvironmentInstance is the interface for the running instance of the environment (container, local etc). Therefore, I had to change the interface itself to add a environment to running instance.

environment/local.go
```diff
@@ -86,6 +86,12 @@ func (e *Local) Stat(name string) (fs.FileInfo, error) {
 	return os.Stat(name)
 }
 
+// AddEnvironmentVariablesToInstance adds the environment variables after the environment is created
+func (e *Local) AddEnvironmentVariablesToInstance(envList []string) error {
+	e.EnvKeyValueList = append(e.EnvKeyValueList, envList...)
+	return nil
+}
```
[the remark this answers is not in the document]
seshapad: see https://github.com/konveyor/move2kube/pull/907#discussion_r1022340116

transformer/external/executabletransformer.go
seshapad: The rationale behind have separate variables is to make it independent of the environment type (container, local etc). Also, 
(1) it allows the added flexibility of defining the I/O for detect and transform scripts separately. 
(2) We could potentially use the same format for the upcoming gRPC invocation method.

environment/environment.go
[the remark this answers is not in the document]
seshapad: done

transformer/external/executabletransformer.go
seshapad: done

transformer/external/executabletransformer.go
seshapad: done

transformer/external/executabletransformer.go
seshapad: shouldnt the user be forced to specify a input modality. Mode has to be mandatory. No?

transformer/external/executabletransformer.go
```diff
@@ -32,20 +32,50 @@ import (
...
+// InputFormat input/output mode definition
+type InputFormat struct {
+	Mode    IOMode `yaml:"mode"`
+	EnvName string `yaml:"envName,omitempty`
+	Path    string
+}
+
+// OutputFormat input/output mode definition
+type OutputFormat struct {
+	Mode    IOMode `yaml:"mode"`
```
ashokponkumar, line 60: Are the modes supported for input and output the same? For example, I am not sure how arg mode or even env mode will work for output

transformer/external/executabletransformer.go
```diff
@@ -32,20 +32,50 @@ import (
...
+}
+
 // ExecutableYamlConfig is the format of executable yaml config
 type ExecutableYamlConfig struct {
 	EnableQA           bool                       `yaml:"enableQA"`
 	Platforms          []string                   `yaml:"platforms"`
-	DirectoryDetectCMD environmenttypes.Command   `yaml:"directoryDetectCMD"`
-	TransformCMD       environmenttypes.Command   `yaml:"transformCMD"`
 	Container          environmenttypes.Container `yaml:"container,omitempty"`
+	DetectCmdFormat    CommandFormat              `yaml:"directoryDetect"`
```
ashokponkumar, line 77: Lets also think through (need not implement grpc in this PR) how the struct should look like if we have to support grpc and design it.

ashokponkumar (changes requested):
Thanks Nabha. Few more comments.

transformer/external/executabletransformer.go
```diff
@@ -32,20 +32,50 @@ import (
...
+	TransformCmdFormat CommandFormat              `yaml:"transform"`
```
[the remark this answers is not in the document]
seshapad: Removed this altogether as discussed.

transformer/external/executabletransformer.go
```diff
@@ -32,20 +32,50 @@ import (
...
+
+// OutputFormat input/output mode definition
+type OutputFormat struct {
+	Mode    IOMode `yaml:"mode"`
+	EnvName string `yaml:"envName,omitempty`
+	Path    string
+}
+
+// CommandFormat input/output mode and command definitions for detect and transform scripts
+type CommandFormat struct {
```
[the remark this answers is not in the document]
seshapad: Removed as discussed.

transformer/external/executabletransformer.go
seshapad: Will be covered in follow-up PR as discussed.

transformer/external/executabletransformer.go
```diff
@@ -204,3 +237,24 @@ func (t *Executable) uploadInput(data interface{}, inputFile string) (string, er
...
+	cmd environmenttypes.Command,
+	kvMap map[string]string) (
+	environmenttypes.Command,
+	[]string,
+) {
+	const dollarPrefix = "$"
+	cmdToRun := cmd
+	for envKey, value := range kvMap {
+		for index, token := range cmdToRun {
+			regexExp := regexp.MustCompile(`\$` + envKey)
```
ashokponkumar, line 251:
```suggestion
			regexExp := regexp.MustCompile(`\$({\()?` + envKey + "(}\))?")
```

Should we allow the brackets too?

ashokponkumar (changes requested):
Thanks ji.. one small suggestion

transformer/external/executabletransformer.go
seshapad: done. The above line had a minor syntax error for the quotes. Hence, I have pushed the change to PR.

seshapad: There seems to a bug in the above regex. Will try to isolate it.

seshapad: done.

ashokponkumar (approved):

PR #912, opened 2022-11-13
fix: IR config was not getting properly unmarshalled leading to empty containers in the deployment yamls

HarikrishnanBalagopal:


ashokponkumar (approved):
Nice work Hari!

PR #923, opened 2022-12-07
fix: look for external services in CfApps file instead of the CfServices file

HarikrishnanBalagopal:
Also change the python code to use the new input and output format.

### DIFF
--- a/assets/built-in/transformers/kubernetes/operatorsfromtca/Dockerfile
+++ b/assets/built-in/transformers/kubernetes/operatorsfromtca/Dockerfile
@@ -1,4 +1,4 @@
 # FROM quay.io/konveyor/tca-service-api
 FROM quay.io/mihirc36/tackle-container-advisor
 
-COPY detect.py transform.py ./
+COPY detect.py transform.py preprocess.py cfservices.yaml ./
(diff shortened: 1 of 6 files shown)


assets/built-in/transformers/kubernetes/operatorsfromtca/Dockerfile
```diff
@@ -1,4 +1,4 @@
 # FROM quay.io/konveyor/tca-service-api
 FROM quay.io/mihirc36/tackle-container-advisor
 
-COPY detect.py transform.py ./
+COPY detect.py transform.py catalogue.yaml ./
```
ashokponkumar, line 4: How can users configure this catalogue?

ashokponkumar (changes requested):
Couple of questions.

assets/built-in/transformers/kubernetes/operatorsfromtca/Dockerfile
HarikrishnanBalagopal: By using a custom transformer. They can replace it with their own catalogue.yaml

ashokponkumar: users should be able to specify a different catalogue for each run, and that should not require changing the custom transformer image.

HarikrishnanBalagopal: done

ashokponkumar (approved):

PR #953, opened 2023-01-26
feat: Added presets for dockerfile and docker-compose configs

seshapad:


assets/built-in/presets/analyze-docker-compose-only.yaml
```diff
@@ -0,0 +1,17 @@
+move2kube:
+  transformers:
+    types:
+      - ComposeGenerator
```
ashokponkumar, line 4: Isn't ComposeAnalyser required?

ashokponkumar (changes requested):

ashokponkumar (approved):

PR #959, opened 2023-02-08
fix: Adds .env file detection in compose

seshapad:

### DIFF
--- a/transformer/compose/v1v2.go
+++ b/transformer/compose/v1v2.go
@@ -93,7 +93,7 @@ func removeNonExistentEnvFilesV2(path string) preprocessFunc {
 					logrus.Debugf("Unable to find env file %s for service %s in file %s. Ignoring it.", envFilePath, serviceName, path)
 				} else {
 					vals[envFile] = defaultEnvFile
-					logrus.Debugf("env config file %s referred in service %s in file %s. Adding  it.", envFilePath, serviceName, path)
+					logrus.Debugf("env file %s found for service %s in file %s. Adding  it.", envFilePath, serviceName, path)
 				}
 			}
 		}
--- a/transformer/compose/v1v2.go
+++ b/transformer/compose/v1v2.go
@@ -95,9 +95,11 @@ func parseV2(path string, interpolate bool) (*project.Project, error) {
 	//TODO: Check if any variable is mandatory
 	someEnvFilePath := ".env"
 	if !common.IgnoreEnvironment {
-		absSomeEnvFilePath, err := filepath.Abs(someEnvFilePath)
+		composeFileDir := filepath.Dir(path)
+		absSomeEnvFilePath := filepath.Join(composeFileDir, someEnvFilePath)
+		_, err := os.Stat(absSomeEnvFilePath)
 		if err != nil {
-			logrus.Errorf("Failed to make the path %s absolute. Error: %q", someEnvFilePath, err)
+			logrus.Errorf("Failed to find the env path %s. Error: %q", someEnvFilePath, err)
 			return nil, err
 		}
 		someEnvFilePath = absSomeEnvFilePath
--- a/transformer/compose/v1v2.go
+++ b/transformer/compose/v1v2.go
@@ -99,7 +99,7 @@ func parseV2(path string, interpolate bool) (*project.Project, error) {
 		absSomeEnvFilePath := filepath.Join(composeFileDir, someEnvFilePath)
 		_, err := os.Stat(absSomeEnvFilePath)
 		if err != nil {
-			logrus.Errorf("Failed to find the env path %s. Error: %q", someEnvFilePath, err)
+			logrus.Debugf("Failed to find the env path %s. Error: %q", someEnvFilePath, err)
 			return nil, err
 		}
 		someEnvFilePath = absSomeEnvFilePath
--- a/transformer/compose/v1v2.go
+++ b/transformer/compose/v1v2.go
@@ -90,7 +90,7 @@ func removeNonExistentEnvFilesV2(path string) preprocessFunc {
 				envFilePath := filepath.Join(composeFileDir, defaultEnvFile)
 				finfo, err := os.Stat(envFilePath)
 				if os.IsNotExist(err) || finfo.IsDir() {
-					logrus.Debugf("Unable to find env config file %s referred in service %s in file %s. Ignoring it.", envFilePath, serviceName, path)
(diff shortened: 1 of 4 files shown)


ashokponkumar (changes requested):

transformer/compose/v3.go
```diff
@@ -80,6 +81,21 @@ func removeNonExistentEnvFilesV3(path string, parsedComposeFile map[string]inter
...
 						}
+					} else {
+						checkForDotEnv = true
+					}
+				} else {
+					checkForDotEnv = true
+				}
+				if checkForDotEnv {
+					vals := val.(map[string]interface{})
+					envFilePath := filepath.Join(composeFileDir, ".env")
```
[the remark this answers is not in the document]
seshapad: done

transformer/compose/v1v2.go
```diff
@@ -95,9 +95,11 @@ func parseV2(path string, interpolate bool) (*project.Project, error) {
 	//TODO: Check if any variable is mandatory
 	someEnvFilePath := ".env"
 	if !common.IgnoreEnvironment {
-		absSomeEnvFilePath, err := filepath.Abs(someEnvFilePath)
+		composeFileDir := filepath.Dir(path)
+		absSomeEnvFilePath := filepath.Join(composeFileDir, someEnvFilePath)
+		_, err := os.Stat(absSomeEnvFilePath)
 		if err != nil {
-			logrus.Errorf("Failed to make the path %s absolute. Error: %q", someEnvFilePath, err)
+			logrus.Errorf("Failed to find the env path %s. Error: %q", someEnvFilePath, err)
```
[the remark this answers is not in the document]
seshapad: done

transformer/compose/v1v2.go
```diff
@@ -93,11 +93,13 @@ func parseV2(path string, interpolate bool) (*project.Project, error) {
...
+	someEnvFilePath := defaultEnvFile
 	if !common.IgnoreEnvironment {
-		absSomeEnvFilePath, err := filepath.Abs(someEnvFilePath)
+		composeFileDir := filepath.Dir(path)
+		absSomeEnvFilePath := filepath.Join(composeFileDir, someEnvFilePath)
+		_, err := os.Stat(absSomeEnvFilePath)
 		if err != nil {
-			logrus.Errorf("Failed to make the path %s absolute. Error: %q", someEnvFilePath, err)
+			logrus.Debugf("Failed to find the env path %s. Error: %q", someEnvFilePath, err)
 			return nil, err
```
ashokponkumar, line 103: Why do we want to return with an error when the .env file is not found?

transformer/compose/v3.go
```diff
@@ -80,6 +81,21 @@ func removeNonExistentEnvFilesV3(path string, parsedComposeFile map[string]inter
 							}
 							vals[envFile] = envfiles
 						}
+					} else {
+						checkForDotEnv = true
+					}
+				} else {
+					checkForDotEnv = true
+				}
+				if checkForDotEnv {
```
ashokponkumar, line 90: Does Docker compose too use the `.env` file only if there is no environment file specified?

transformer/compose/v3.go
```diff
@@ -80,6 +81,21 @@ func removeNonExistentEnvFilesV3(path string, parsedComposeFile map[string]inter
...
+					vals := val.(map[string]interface{})
+					envFilePath := filepath.Join(composeFileDir, defaultEnvFile)
+					finfo, err := os.Stat(envFilePath)
+					if os.IsNotExist(err) || finfo.IsDir() {
+						logrus.Debugf("Unable to find env config file %s referred in service %s in file %s. Ignoring it.", envFilePath, serviceName, path)
```
ashokponkumar, line 95: If we are here, it means it is not referenced in the docker compose file, isn't it?

transformer/compose/v3.go
```diff
@@ -80,6 +81,21 @@ func removeNonExistentEnvFilesV3(path string, parsedComposeFile map[string]inter
...
+					} else {
+						vals[envFile] = defaultEnvFile
+						logrus.Debugf("env config file %s referred in service %s in file %s. Adding  it.", envFilePath, serviceName, path)
```
ashokponkumar, line 98: If we are here, it means it is not referenced in the docker compose file, isn't it?

ashokponkumar (changes requested):

transformer/compose/v3.go
seshapad: if we hit this condition, it means that docker compose has not listed any env file. So, as a follow-on, we also check if .env is available or not. If available we add it to the env list, if the file does not exist, then we dont add anything.

transformer/compose/v3.go
seshapad: if we hit this condition, it means that docker compose has not listed any env file. So, as a follow-on, we also check if .env is available or not. If available we add it to the env list, if the file does not exist, then we dont add anything.

transformer/compose/v1v2.go
seshapad: Removed this return.

transformer/compose/v3.go
ashokponkumar, line 95:
```suggestion
						logrus.Debugf("Unable to find env file %s for service %s in file %s. Ignoring it.", envFilePath, serviceName, path)
```

transformer/compose/v3.go
ashokponkumar, line 98:
```suggestion
						logrus.Debugf("env file %s for service %s in file %s. Adding  it.", envFilePath, serviceName, path)
```

ashokponkumar (changes requested):

transformer/compose/v1v2.go
```diff
@@ -93,12 +108,13 @@ func parseV2(path string, interpolate bool) (*project.Project, error) {
...
-		absSomeEnvFilePath, err := filepath.Abs(someEnvFilePath)
+		composeFileDir := filepath.Dir(path)
+		absSomeEnvFilePath := filepath.Join(composeFileDir, someEnvFilePath)
+		_, err := os.Stat(absSomeEnvFilePath)
 		if err != nil {
-			logrus.Errorf("Failed to make the path %s absolute. Error: %q", someEnvFilePath, err)
-			return nil, err
+			logrus.Debugf("Failed to find the env path %s. Error: %q", someEnvFilePath, err)
 		}
 		someEnvFilePath = absSomeEnvFilePath
```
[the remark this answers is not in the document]
seshapad: Done.

ashokponkumar (approved):

PR #969, opened 2023-02-19
fix: set-output throws error in ci/cd starting June 2023

venkatbandarupalli:
This is required as set-output would be completely removed and the workflows can potentially break starting June 2023.

Fyi - 
I am new to open source, however, I have some working knowledge on git and enterprise github.

I tried following the conventions as much as possible, please guide me if I have missed anything in this pull request.

thank you.


HarikrishnanBalagopal (approved):
LGTM

Issue #978, opened 2023-02-24
Add fix to resolve issues with docker secrets defined with special characters


mnarkhede85:
**Describe the bug**
M2K tool is not replacing special characters used in Docker secret after transforming to Kubernetes/Openshift manifests files.

**To Reproduce**
1. Define and create the docker secret with special character ( -, _ , . ) etc.  Ex - test_secret , test_secret.properties 
2. Run move2kube on docker-compose file referencing to docker secret 
3. Verify the content and name of secret yaml file created.
4. Run Kubectl apply to see the error.

The Secret "test_options" is invalid: metadata.name: Invalid value: "test_options": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

**Expected behavior**
The secret yaml should be updated with kubenetes supported RFC naming standards.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. RHEL]
 - Version [e.g. 22]. `move2kube version` should return it.

**Additional context**

PR #988, opened 2023-03-09
fix: Normalized the secret and configmap names in v3 compose

seshapad:


ashokponkumar:
Do we need to do this for v1v2 too?

ashokponkumar:
Fixes #978

Issue #1004, opened 2023-04-03
clean subcommand or sanity cleanup before command run of temporary directory


kmehant:
**Is your feature request related to a problem? Please describe.**
In the current move2kube behavior, if a certain move2kube subcommand flow is interrupted it leaves the temporary residue as is.

**Describe the solution you'd like**
There are two solutions we can implement
1. add a new subcommand clean which can be used by the user to clean the clutter
2. do sanity or clean-up work before any subcommand run.

We might need to decide which one to choose.

**Describe alternatives you've considered**
Manually deleting the temp folders

kmehant:
@ashokponkumar I can take this up.

kmehant:
Related - https://github.com/konveyor/move2kube/issues/860#issue-1362709213

Issue #860, opened 2022-09-06
Cache at /tmp/ persists when interrupting the command / terminating the process


kmehant:
**Describe the bug**
Cache is not being cleared when the command is interruped / terminating the process

**To Reproduce**
Start a transformation
```
move2kube transform -s source/folder
```
Interrupt the command.
```
command  + C
```
Cache still persists
```
ls /tmp/ | grep move2kube
```

**Expected behavior**
Cache should be cleared

**Screenshots**
<img width="987" alt="Screenshot 2022-09-06 at 9 54 36 AM" src="https://user-images.githubusercontent.com/15800200/188550979-81072fcc-8cee-4b81-a74e-412e97b18034.png">

**Desktop (please complete the following information):**
 - OS: macOS Monterey Version 12.5.1
 - Version [e.g. 22]. v0.3.0+unreleased

PR #1008, opened 2023-04-04
fix: remove temp directories on interrupt

kmehant:


ashokponkumar (approved):

kmehant:
fixes https://github.com/konveyor/move2kube/issues/1004 https://github.com/konveyor/move2kube/issues/860

PR #1018, opened 2023-04-07
feat: Optimize move2kube collect performance using ListAppsByQuery with depth 0

Akash-Nayak:


ashokponkumar:
If we do this, then the data available in the collect will be less, which is not usable for transform then.

Akash-Nayak:
Closing this PR as we have a new PR https://github.com/konveyor/move2kube/pull/1026, which enhances the `move2kube collect` functionality by using a hybrid approach to collect only the selected CF apps metadata through an input yaml, and also supports query depth and spaceguid based filtering.

PR #1019, opened 2023-04-10
fix: remove external transformer temp directories on interrupt

kmehant:
Signed-off-by: Mehant Kammakomati <mehant.kammakomati2@ibm.com>

### DIFF
--- a/lib/transformer.go
+++ b/lib/transformer.go
@@ -25,6 +25,7 @@ import (
 	"github.com/konveyor/move2kube/common"
 	"github.com/konveyor/move2kube/qaengine"
 	"github.com/konveyor/move2kube/transformer"
+	"github.com/konveyor/move2kube/transformer/external"
 	plantypes "github.com/konveyor/move2kube/types/plan"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


ashokponkumar:
@seshapad are these paths used inside the container or external to them?

kmehant:
Screenshot 2023-04-10 at 11 01 00 PM

@ashokponkumar @seshapad they are used externally.

ashokponkumar:
Hmm.. how does it work in windows then.. there might be a larger issue in play here then..

kmehant:
https://github.com/konveyor/move2kube/blob/73ab28607a7350ce5b75cc3afcfb6487eb059d3d/environment/local.go#L139-L156

this function seems to do this. 

@seshapad can confirm.

kmehant:
@ashokponkumar https://github.com/konveyor/move2kube/pull/1019#issuecomment-1502094717

Yes, there should be a drive and the way paths are joined would different.

seshapad:
> @seshapad are these paths used inside the container or external to them?

Yes @ashokponkumar . It's meant to let configure the external transformer on where to pick input and where to write output.

PR #1027, opened 2023-04-17
feat: allow git remote paths for move2kube generated outputs 

venkatbandarupalli:


lib/transformer.go
```diff
@@ -49,11 +50,17 @@ func Transform(ctx context.Context, plan plantypes.Plan, preExistingPlan bool, o
 	requirements, _ := selectorsInPlan.Requirements()
 	transformerSelectorObj = transformerSelectorObj.Add(requirements...)
 
+	remoteOutputFSPath := vcs.GetInputClonedPath(outputPath, common.RemoteOutputsFolder, true)
```
kmehant, line 53: Can you change the function name `GetInputClonedPath` to something relevant to the usecase as I see this is being used widely now. May be `GetClonedPath` or something?

venkatbandarupalli: Hello @kmehant I have addressed this in https://github.com/konveyor/move2kube/pull/1031

kmehant:
@venkatbandarupalli can you rebase this?

kmehant (changes requested):
@venkatbandarupalli good work, A couple of changes

1. What if the branch the user provides does not exist in the remote, will the current handle that? if not, please change it so it creates a new branch if not present
2. on the basis of the flag `ignore-env` already being used in the move2kube code, can you pick up git credentials accordingly, rather than asking the user on compulsion?

venkatbandarupalli:
@kmehant Thank you for your suggestions.

1. I have fixed this.
2. The go-git package does not support pulling global git credentials configured on the system environment. I have got this information from the official go-git compatibility log (https://github.com/go-git/go-git/blob/master/COMPATIBILITY.md). For now, I have added we do not support this warning. We can perhaps work on this on subsequent pull requests.

kmehant (approved):
LGTM

PR #1029, opened 2023-04-26
test: add unit tests for ir.go file

tarun8718:
Tests for ir.go file in ir package
Screenshot 2023-04-26 at 3 43 05 PM


types/ir/ir_test.go
```diff
@@ -0,0 +1,287 @@
...
+		newIR.Name = "sampleIR"
+		newIR.AddService(NewServiceWithName("sampleSerivce"))
+		oldIR.Merge(newIR)
+		if !reflect.DeepEqual(oldIR, newIR) {
+			t.Errorf("failed to merge new IR into old IR. Old IR : %+v, New IR: %+v", oldIR, newIR)
+		}
+	})
+
+	t.Run("test for merging containers", func(t *testing.T) {
+		oldContainer := NewContainer()
```
HarikrishnanBalagopal, line 256: Let's add more comprehensive tests instead of using empty objects.
Most of the merge functionality will be skipped if the fields in the object are empty/nil.

We can `logrus.Infof("%#v", ir)` in the `ir.go` file and run a proper `move2kube transform -s language-platforms` to get a nearly full IR. Will be closer to real world use.

HarikrishnanBalagopal (changes requested):

types/ir/ir_test.go
tarun8718: I've addressed this in the new commits. Please review, Thanks!

HarikrishnanBalagopal (approved):

PR #1037, opened 2023-05-28
fix: follow git branch naming convention based on docs.github.com

kmehant:
fixed based on https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags

### DIFF
--- a/common/vcs/git.go
+++ b/common/vcs/git.go
@@ -53,7 +53,7 @@ func isGitCommitHash(commithash string) bool {
 }
 
 func isGitBranch(branch string) bool {
-	gitBranchRegex := regexp.MustCompile(`^(main|development|master|(features|tests|(bug|hot)fix)((\/|-)[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*){1,2}|release(\/|-)[0-9]+(\.[0-9]+)*(-(alpha|beta|rc)[0-9]*)?)$`)
+	gitBranchRegex := regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\.\-_\/]*$`)
 	return gitBranchRegex.MatchString(branch)
 }
 


common/vcs/git.go
seshapad, line 1: @kmehant are these `main|development|master|(features|tests|(bug|hot)fix` the only branch prefix names? Is there any standard around this?
cc: @ashokponkumar

kmehant: The current fix makes it generic and opens up possibilities and also taking care of the breaking special characters out of branch names.

seshapad: @kmehant if you are fixing the format of the branch to specific names in the prefix which are not necessarily part of the standard, would it not affect generality?
Would it be better to let the prefix be any name which is not part of this set `main|development|master|(features|tests|(bug|hot)fix`?

common/vcs/git.go
```diff
@@ -53,7 +53,7 @@ func isGitCommitHash(commithash string) bool {
 }
 
 func isGitBranch(branch string) bool {
-	gitBranchRegex := regexp.MustCompile(`^(main|development|master|(features|tests|(bug|hot)fix)((\/|-)[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*){1,2}|release(\/|-)[0-9]+(\.[0-9]+)*(-(alpha|beta|rc)[0-9]*)?)$`)
+	gitBranchRegex := regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9\.\-_\/]*$`)
```
kmehant, line 56: @seshapad The fix is this!

common/vcs/git.go
kmehant: https://github.com/konveyor/move2kube/pull/1037/files#r1209033547


> @kmehant if you are fixing the format of the branch to specific names in the prefix which are not necessarily part of the standard, would it not affect generality?


@seshapad ^ yes that's the reason we have this fix.

ashokponkumar (approved):

Issue #881, opened 2022-09-28
test: Test the filesystem package.


HarikrishnanBalagopal:
> ⚠️  **Don't forget to read our contribution guidelines before you start working on an issue** https://github.com/konveyor/move2kube/blob/main/contributing.md

# Description

Since Move2Kube moved from v1, v2 and to v3 architecture the test coverage has dropped significantly.
A good candidate for unit tests are the files in the `filesystem` package.
It is a core part of Move2Kube's `transform` functionality and yet has no unit tests.

https://github.com/konveyor/move2kube/tree/ccc92502e41d67756929f36a9f7c0938bfb58c78/filesystem

This package deals with generating and manipulating files and directories.
It can copy files/directories from a source to a destination.
It can also fill template files before they get copied to their destination.
While copying directories It can also calculate the difference between the source and destination directory and only copy the minimal number of files required.

# How to get started

For this issue you should start with some simple tests that creates temporary files and directories and calls the functions on it.
Next step would be to make slight changes to the files in the directory and create a separate subtest for each scenario.
Some scenarios that are good to test:
- Copying a normal file
- Copying a template file
- Copying a directory full of template files
- Function called on an empty directory.
- Directory we don't have permission to read.
- Directory with several subdirectories

# How to add unit tests

Some guidelines:
- Look at the other tests in the package/project and write similar tests. Example:
  https://github.com/konveyor/move2kube/blob/main/types/collection/cluster_test.go#L28-L105
- Use subtests to test different paths through the function.
- Don't worry about testing every single path through the function. Focus on common use cases and failure modes.

Some helpful resources on how to write unit tests in Go:
- https://blog.golang.org/subtests
- https://youtu.be/hVFEV-ieeew

# Code to be tested

https://github.com/konveyor/move2kube/tree/ccc92502e41d67756929f36a9f7c0938bfb58c78/filesystem

ritesh-karankal:
Hi @HarikrishnanBalagopal,
Can I work on this issue?

HarikrishnanBalagopal:
> Hi @HarikrishnanBalagopal, Can I work on this issue?

Sure

ritesh-karankal:
Hi @HarikrishnanBalagopal 
I am a beginner in golang. I expected to learn and complete the task, but  I guess I won't be able to resolve this issue.
I don't want to keep this issue on hold for long.
Sorry for the inconvenience, but I'd like to unassign myself.
I look forward to contributing to future issues.

[...]

PR #1066, opened 2023-07-28
test: add tests for replicateProcessFileCallBack

Prakhar-Agarwal-byte:
Added tests for replicateProcessFileCallBack in the filesystem package. These tests cover 4 cases
1. Source file does not exist, should return an error
2. Destination file doesn't exist, should copy the source file
3. Destination file exists but is different, should copy the source file
4. Destination file exists and is the same, should return nil without copying

Fixes a part of #881

### DIFF
--- a/filesystem/replicator_test.go
+++ b/filesystem/replicator_test.go
@@ -1,121 +1,141 @@
+/*
+ *  Copyright IBM Corporation 2021
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
 package filesystem
 
 import (
+	"bytes"
+	"io/ioutil"
 	"os"
 	"testing"
-	"time"
 )
-
-func TestReplicateProcessFileCallBack(t *testing.T) {
-	t.Run("NonExistentSourceFile", func(t *testing.T) {
-		sourceFilePath := "non_existent_file.txt"
-		destinationFilePath := "destination.txt"
-		config := struct{}{}
-		err := replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
-		if err == nil {
-			t.Errorf("Expected error for non-existent source file, got nil")
-		}
-	})
-
-	t.Run("CopySourceToDestination", func(t *testing.T) {
-		sourceFile, err := os.Create("source_file.txt")
-		if err != nil {
-			t.Fatal(err)
-		}
-		defer func() {
-			sourceFile.Close()
-			os.Remove("source_file.txt")
-		}()
-
-		sourceFilePath := "source_file.txt"
-		destinationFilePath := "destination_file.txt"
-		config := struct{}{}
-
-		err = replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
-		if err != nil {
-			t.Errorf("Expected nil error, got: %s", err)
-		}
-	})
-
-	t.Run("CopySourceToDestinationDifferent", func(t *testing.T) {
-		sourceFile, err := os.Create("source_file.txt")
-		if err != nil {
-			t.Fatal(err)
-		}
-		defer func() {
-			sourceFile.Close()
-			os.Remove("source_file.txt")
-		}()
-
-		_, err = sourceFile.WriteString("hello, world!")
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		sourceFile.Close() // Close and reopen the file to update the modification time
-		time.Sleep(1 * time.Second)
-
-		destinationFile, err := os.Create("destination_file.txt")
-		if err != nil {
-			t.Fatal(err)
-		}
-		defer func() {
-			destinationFile.Close()
-			os.Remove("destination_file.txt")
-		}()
-
-		_, err = destinationFile.WriteString("hello, go!")
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		sourceFilePath := "source_file.txt"
-		destinationFilePath := "destination_file.txt"
-		config := struct{}{}
-
-		err = replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
-		if err != nil {
-			t.Errorf("Expected nil error, got: %s", err)
-		}
-	})
-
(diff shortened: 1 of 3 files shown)


filesystem/replicator_test.go
```diff
@@ -0,0 +1,121 @@
...
+		destinationFilePath := "destination.txt"
+		config := struct{}{}
+		err := replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
+		if err == nil {
+			t.Errorf("Expected error for non-existent source file, got nil")
+		}
+	})
+
+	t.Run("CopySourceToDestination", func(t *testing.T) {
+		sourceFile, err := os.Create("source_file.txt")
```
HarikrishnanBalagopal, line 21: Instead of creating an empty file we can put a file under `testdata` folder and use that. It also mean we don't need to delete the file, so can remove the `defer os.Remove()`

filesystem/replicator_test.go
```diff
@@ -0,0 +1,121 @@
...
+		}()
+
+		sourceFilePath := "source_file.txt"
+		destinationFilePath := "destination_file.txt"
+		config := struct{}{}
+
+		err = replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
+		if err != nil {
+			t.Errorf("Expected nil error, got: %s", err)
+		}
```
HarikrishnanBalagopal, line 37: Also we can check the contents of the copied file against the source file contents.

filesystem/replicator_test.go
```diff
@@ -0,0 +1,121 @@
...
+			sourceFile.Close()
+			os.Remove("source_file.txt")
+		}()
+
+		_, err = sourceFile.WriteString("hello, world!")
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		sourceFile.Close() // Close and reopen the file to update the modification time
```
HarikrishnanBalagopal, line 55: Can't we change the file's "last modification time" by  appending to the file? Let's try to avoid `time.Sleep` if we can in our tests

HarikrishnanBalagopal (changes requested):
Some minor changes

filesystem/replicator_test.go
```diff
@@ -0,0 +1,121 @@
+package filesystem
```
kmehant, line 1: Please, add the IBM license header.
You can pick it up from one of the existing code files.

filesystem/replicator_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"os"
+	"testing"
+	"time"
+)
+
+func TestReplicateProcessFileCallBack(t *testing.T) {
+	t.Run("NonExistentSourceFile", func(t *testing.T) {
```
kmehant, line 10: `NonExistentSourceFile` Shall we have a proper test title? What the test case is about?

kmehant: Same goes for other test cases as well.

kmehant (changes requested):

filesystem/replicator_test.go
```diff
@@ -1,121 +1,141 @@
...
+		 destinationFilePath := "test_data/replicator_test/destination.txt"
+		 config := struct{}{}
+		 err := replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
+		 if err == nil {
+			 t.Errorf("Expected error for non-existent source file, got nil")
+		 }
+	 })
+ 
+	 t.Run("test for destination file doesn't exist, should copy the source file", func(t *testing.T) {
+		 sourceFilePath := "test_data/replicator_test/emptyfile.txt"
```
HarikrishnanBalagopal, line 38: I think the folder name has to be `testdata` https://github.com/golang/go/issues/14715
Also helps consistency https://github.com/konveyor/move2kube/tree/main/common/testdata

filesystem/replicator_test.go
```diff
@@ -1,121 +1,141 @@
...
+			 t.Errorf("Expected destination content to be equal to source content, but they are different.")
+		 }
+	 })
+ 
+	 t.Run("test for destination file exists and is the same, should return nil without copying", func(t *testing.T) {
+		 sourceFilePath := "test_data/replicator_test/source_file.txt"
+		 destinationFilePath := "test_data/replicator_test/destination_file.txt"
+		 config := struct{}{}
+		 
+		 sourceFile, err := os.Create(sourceFilePath)
```
HarikrishnanBalagopal, line 107: Why do we need to create these files? We can have both files in the `testdata` folder.

filesystem/replicator_test.go
```diff
@@ -1,121 +1,141 @@
...
+		 _, err = destinationFile.WriteString("hello, world!")
+		 if err != nil {
+			 t.Fatal(err)
+		 }
+ 
+		 err = replicateProcessFileCallBack(sourceFilePath, destinationFilePath, config)
+		 if err != nil {
+			 t.Errorf("Expected nil error, got: %s", err)
+		 }
+	 })
```
HarikrishnanBalagopal, line 139: We need to check that the destination file exists and its contents hasn't been changed after the `replicateProcessFileCallBack` function call.

filesystem/replicator_test.go
```diff
@@ -1,121 +1,141 @@
...
-			t.Errorf("Expected nil error, got: %s", err)
-		}
-	})
-}
+ 
+ func TestReplicateProcessFileCallBack(t *testing.T) {
+	 t.Run("test for source file does not exist, should return an error", func(t *testing.T) {
+		 sourceFilePath := "test_data/replicator_test/non_existent_file.txt"
+		 destinationFilePath := "test_data/replicator_test/destination.txt"
+		 config := struct{}{}
```
HarikrishnanBalagopal, line 30: we can give `nil` since the `config` parameter has type `interface{}` .

HarikrishnanBalagopal (changes requested):
a few minor things

filesystem/replicator_test.go
Prakhar-Agarwal-byte: There's one folder already present in the filesystem package with name "test_data". Shall I change it to "testdata"?

Prakhar-Agarwal-byte: https://github.com/konveyor/move2kube/tree/main/filesystem/test_data

HarikrishnanBalagopal: yes

Prakhar-Agarwal-byte: I have added the test files for this PR in "testdata" folder. Renaming the "test_data" folder will introduce changes in other files too, which may not be related to the exact purpose of this PR. Shall I can raise a separate PR for renaming it?

HarikrishnanBalagopal (approved):
LGTM

kmehant (approved):
LGTM

filesystem/replicator_test.go
Prakhar-Agarwal-byte: @HarikrishnanBalagopal I have created the [PR ](https://github.com/konveyor/move2kube/pull/1072)for renaming the `test_data` folder to `testdata`

PR #1070, opened 2023-08-02
fix: Adds an input recieveing option to recieve input for argoCD transformer namespace.

### LINKED ISSUE #930: bug: in the ArgoCD application yaml actually set the namespace based on user input
# Bug

## Overview

Currently the field is not being set by anything.
https://github.com/konveyor/move2kube/blob/8aa892e2b4b50eaa67073358dbc85f6988dd7f3c/transformer/kubernetes/apiresource/argocdapplication.go#L84

https://github.com/konveyor/move2kube/blob/8aa892e2b4b50eaa67073358dbc85f6988dd7f3c/types/ir/argocd.go#L31

The field `irApplication.DestNamespace` is only referenced in 2 places and it's not being set by either of them.
image

## Fix

Probably we can use ask the user to provide the namespace where the app should be deployed once the CI/CD pipeline finishes.

MdSahil-oss:
This PR adds an input recieveing option to recieve input for argoCD transformer namespace.

This PR fixes #930.


transformer/kubernetes/argocdtransformer.go
```diff
@@ -152,8 +153,26 @@ func (t *ArgoCD) setupEnhancedIR(oldir irtypes.IR, projectName string) irtypes.E
 		return common.MakeStringDNSSubdomainNameCompliant(fmt.Sprintf("%s-%s", projectName, baseName))
 	}
 	appName := p(baseAppName)
-	ir.ArgoCDResources = irtypes.ArgoCDResources{
-		Applications: []irtypes.Application{{Name: appName}},
+	destNamespace := qaengine.FetchStringAnswer(
+		common.ConfigTransformersKubernetesArgocdNamespaceKey,
+		"Enter the namespace where argocd pipelibe to be deployed",
```
HarikrishnanBalagopal, line 158: typo `pipelibe`

MdSahil-oss: @HarikrishnanBalagopal fixed it.

transformer/kubernetes/argocdtransformer.go
```diff
@@ -152,8 +153,26 @@ func (t *ArgoCD) setupEnhancedIR(oldir irtypes.IR, projectName string) irtypes.E
...
+				Name: appName,
+			}},
+		}
+	} else {
+		ir.ArgoCDResources = irtypes.ArgoCDResources{
+			Applications: []irtypes.Application{{
+				Name:          appName,
+				DestNamespace: destNamespace,
+			}},
+		}
```
HarikrishnanBalagopal, line 175: we don't need an if/else since in Golang the default for a variable/struct field `DestNamespace` of type `string` is the empty string  `""`

HarikrishnanBalagopal (changes requested):

transformer/kubernetes/argocdtransformer.go
MdSahil-oss: @HarikrishnanBalagopal done 👍🏽

PR #1076, opened 2023-08-14
fix: Fix install script on Apple devices with Apple Silicon

Soumil-07:
On Apple devices with Apple Silicon, such as my M1 MacBook, the install script fails as it tries to install linux version of jq.


scripts/install.sh
```diff
@@ -164,7 +164,7 @@ installDependencies() {
     JQ='jq'
     if [ "$HAS_JQ" = 'false' ]; then
         JQ="$(mktemp -d)/jq"
-        if [ "${OS}-${ARCH}" = 'darwin-amd64' ]; then
+        if [ "${OS}-${ARCH}" = 'darwin-amd64' ] || [ "${OS}-${ARCH}" = 'darwin-arm64' ]; then
             download 'https://github.com/stedolan/jq/releases/download/jq-1.6/jq-osx-amd64' "$JQ"
```
HarikrishnanBalagopal, line 168: this installs the amd64 version which is not suitable for arm64 M1, M2 chips

HarikrishnanBalagopal (changes requested):

scripts/install.sh
HarikrishnanBalagopal: The 1.6 stable version doesn't have darwin arm64 binary but the newer 1.7rc1 prerelease does
https://github.com/jqlang/jq/releases/download/jq-1.7rc1/jq-macos-arm64
We can switch over if there are not too many breaking changes

Soumil-07: I used the same darwin-amd64 binary as there wasn't a stable build for ARM Macs. The amd64 binary works fine with Rosetta emulation on my machine, should I switch over to the RC release?

Soumil-07: @HarikrishnanBalagopal I've added a commit downloading jq v1.7rc for darwin-arm64. I've checked the release notes and the changes in v1.7 shouldn't affect the installation script.

HarikrishnanBalagopal (approved):
LGTM

PR #1077, opened 2023-08-18
test: add tests for replicateDeletionCallback

Prakhar-Agarwal-byte:
Added test for replicateDeletionCallback in the filesystem package. These tests cover 2 cases

1. Destination folder gets created
2. Created destination folder has the same permissions as the source.

Fixes a part of https://github.com/konveyor/move2kube/issues/881

### DIFF
--- a/filesystem/replicator_test.go
+++ b/filesystem/replicator_test.go
@@ -136,23 +136,7 @@ func TestReplicateProcessFileCallBack(t *testing.T) {
 	 })
  }
  func TestReplicateDeletionCallBack(t *testing.T) {
-		t.Run("test for destination folder gets created", func(t *testing.T) {
-			sourceDir := "testdata/src"
-			destDir := "testdata/dest"
-
-			err := replicateDeletionCallBack(sourceDir, destDir, nil)
-			if err != nil {
-				t.Errorf("Expected no error, but got: %v", err)
-			}
-			defer os.RemoveAll(destDir)
-			// Check if the destination directory exist
-			_, err = os.Stat(destDir)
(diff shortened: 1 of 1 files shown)


seshapad:
@Prakhar-Agarwal-byte Seems like some automated test has failed. Can you please check and fix your code to address it.

filesystem/replicator_test.go
```diff
@@ -136,4 +136,45 @@ import (
...
+			defer os.RemoveAll(destDir)
+			// Check if the destination directory exist
+			_, err = os.Stat(destDir)
+			if err != nil {
+				t.Errorf("Destination directory not created: %v", err)
+			}
+		})
+
+		t.Run("test for permission preservation", func(t *testing.T) {
+			sourceDir := "testdata/src"
```
HarikrishnanBalagopal, line 157: The source doesn't exist. It needs to exist so that the destination folder can be created with the same permissions.

filesystem/replicator_test.go
```diff
@@ -136,4 +136,45 @@ import (
 		}
 	 })
  }
+ func TestReplicateDeletionCallBack(t *testing.T) {
+		t.Run("test for destination folder gets created", func(t *testing.T) {
```
HarikrishnanBalagopal, line 140: We can remove this 1st test and just keep the 2nd test that tests for both creation and permission.

HarikrishnanBalagopal (changes requested):

HarikrishnanBalagopal (approved):
LGTM

PR #1069, opened 2023-08-01
fix: take destNamespace from user

SVKREP:
I am new to open source. I made the PR referring to this link: https://github.com/konveyor/move2kube/blob/main/contributing.md Please LMK if I need to correct something. 

This PR Solves: https://github.com/konveyor/move2kube/issues/930

I've built and tried Move2kube project. I am having trouble in testing this change. Can somebody suggest. Thanks!

PR #1081, opened 2023-08-26
fix: PR930 ArgoCD pipeline set the namespace based on user input

zyy98:
Link to bug: https://github.com/konveyor/move2kube/issues/930

This PR gets the namespace based on user input.

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -155,6 +155,8 @@ const (
 	ConfigCICDTektonGitRepoBasicAuthSecretNameKey = ConfigCICDTektonKey + d + "gitrepobasicauthsecret"
 	// ConfigCICDTektonRegistryPushSecretNameKey is for Tekton push image to registry credentials
 	ConfigCICDTektonRegistryPushSecretNameKey = ConfigCICDTektonKey + d + "registrypushsecret"
+	//ConfigCICDArgoDestinationNameSpace gives the user input namespace destination
+	ConfigCICDArgoDestinationNameSpace = "destnamespace"
 	//ConfigTargetExistingVersionUpdate represents key which how to update versions
 	ConfigTargetExistingVersionUpdate = ConfigTargetKey + d + "existingversionupdate"
 	//ConfigImageRegistryURLKey represents image registry url Key
--- a/transformer/kubernetes/apiresource/argocdapplication.go
+++ b/transformer/kubernetes/apiresource/argocdapplication.go
@@ -19,6 +19,7 @@ package apiresource
 import (
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
 	"github.com/konveyor/move2kube/common"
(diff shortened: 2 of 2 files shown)


common/constants.go
```diff
@@ -155,6 +155,8 @@ const (
 	ConfigCICDTektonGitRepoBasicAuthSecretNameKey = ConfigCICDTektonKey + d + "gitrepobasicauthsecret"
 	// ConfigCICDTektonRegistryPushSecretNameKey is for Tekton push image to registry credentials
 	ConfigCICDTektonRegistryPushSecretNameKey = ConfigCICDTektonKey + d + "registrypushsecret"
+	//ConfigCICDArgoDestinationNameSpace gives the user input namespace destination
+	ConfigCICDArgoDestinationNameSpace = "destnamespace"
```
Akash-Nayak, line 159: Can you change the variable name from `ConfigCICDArgoDestinationNameSpace` to `ConfigCICDArgoCDDestinationNameSpaceKey`?

Also, can you add a new constant`ConfigCICDArgoCDKey = ConfigCICDKey + d + "argocd"` (like we have for Tekton [here](https://github.com/konveyor/move2kube/blob/18bb00d8c08885e62c24c33a704f506d7f51f771/common/constants.go#L151)), and change the value of `ConfigCICDArgoCDDestinationNameSpaceKey` to `ConfigCICDArgoCDKey + d + "destnamespace"`

transformer/kubernetes/apiresource/argocdapplication.go
```diff
@@ -70,6 +71,16 @@ func (*ArgoCDApplication) createNewResource(irApplication irtypes.Application, t
 		clusterServer = deployToSameCluster
 	}
 	appGVK := v1alpha1.ApplicationSchemaGroupVersionKind
+	destNamespace := qaengine.FetchStringAnswer(
+		common.ConfigCICDArgoDestinationNameSpace,
+		"Enter the destination namespace for argo cd pipeline",
```
Akash-Nayak, line 76: Can we change the string to "Enter the destination namespace for the Argo CD pipeline"

transformer/kubernetes/apiresource/argocdapplication.go
```diff
@@ -70,6 +71,16 @@ func (*ArgoCDApplication) createNewResource(irApplication irtypes.Application, t
...
+		[]string{},
```
Akash-Nayak, line 77: Instead of providing empty context `[]string{}`, can we provide a context something like `[]string{If Argo CD pipeline is not relevant to you, then leave empty to use the default value for it.}`

Akash-Nayak (changes requested):
Thanks for your first PR contribution to Move2Kube! A few suggestions..

HarikrishnanBalagopal:
Fixed by #1069

PR #1087, opened 2023-09-06
fix: panic when a port is not specified in an Openshift Route

HarikrishnanBalagopal:


ashokponkumar:
If port is nil, which port are we exposing?

HarikrishnanBalagopal:
> If port is nil, which port are we exposing?

```
	// If specified, the port to be used by the router. Most routers will use all
	// endpoints exposed by the service by default - set this value to instruct routers
	// which port to use.
	Port *RoutePort `json:"port,omitempty" protobuf:"bytes,5,opt,name=port"`
```

seshapad (approved):

ashokponkumar (approved):

PR #1094, opened 2023-09-28
fix: the logic to accept a user provided PEM formatted SSH private key for each git repo domain

HarikrishnanBalagopal:


common/sshkeys/sshkeys.go
```diff
@@ -107,65 +107,56 @@ func loadSSHKeysOfCurrentUser() {
 
 	// Ask whether to load private keys or provide own key
 	options := []string{
-		"Load private ssh keys from " + privateKeyDir,
+		fmt.Sprintf("Load the private SSH keys from the directory '%s'" + privateKeyDir),
```
seshapad, line 110: Where is the privateKeyDir available? Is it provided as input or do we pick up `.ssh` dir on the client machine?

common/sshkeys/sshkeys.go
```diff
@@ -218,61 +202,70 @@ func loadSSHKey(filename string) (string, error) {
...
 func GetSSHKey(domain string) (string, bool) {
 	loadSSHKeysOfCurrentUser()
 	if len(privateKeysToConsider) == 0 {
 		return "", false
 	}
-	if privateKeysToConsider[0] == userProvidedKey {
-		key := qaengine.FetchStringAnswer(common.ConfigRepoPrivKey, "Provide your own PEM-formatted private key:", []string{"Should not be empty"}, "", nil)
+
+	if len(privateKeysToConsider) == 1 && privateKeysToConsider[0] == shoudlAskUserForSSHKey {
+		qaKey := common.JoinQASubKeys(common.ConfigRepoKeysKey, `"`+domain+`"`, "keyData")
```
seshapad, line 226: Where do we get the domain from?

HarikrishnanBalagopal: it's given as input to the function

seshapad: I meany where does move2kube get the domain in the first place. Is it an user input?

common/sshkeys/sshkeys.go
seshapad: Does this function read the `.ssh` directory of the user?

common/sshkeys/sshkeys.go
HarikrishnanBalagopal: `I meant where does move2kube get the domain in the first place. Is it an user input?` from the `git remote -v` output I think

seshapad: Got it. Thanks.

seshapad (approved):

PR #1101, opened 2023-10-26
feat: use podman for building custom images if executable is found

Soumil-07:


common/utils.go
```diff
@@ -1565,3 +1566,12 @@ func IsHTTPURL(str string) bool {
 
 	return regex.MatchString(str)
 }
+
+// GetDefaultContainerRuntime returns the preferred container runtime of the host
+func GetDefaultContainerRuntime() string {
+	_, err := exec.LookPath("podman")
+	if err == nil {
```
Akash-Nayak, line 1573: Can we log a message at level info here that "Podman executable found in the directories named by the PATH environment variable. Using Podman as the container runtime.

shashank381:
Raised a new PR for this feat. https://github.com/konveyor/move2kube/pull/1157

Work done in this PR:

- Added multiarch support using podman.
- Addressed above review comments in https://github.com/konveyor/move2kube/pull/1157.
- Tested the contribution from this PR.

cc: @Akash-Nayak

PR #1102, opened 2023-11-02
feat: warn user if custom directory contains no transformer manifests

Soumil-07:
a common problem that came up for me while debugging custom transformers was adding the Starlark files to my customization directory, but forgetting the manifest for it. This silently failed to load my custom transformers, leaving me scratching my head for hours. 

This PR adds a helpful warning message if that ever happens to anyone else:
image


HarikrishnanBalagopal (approved):
LGTM

PR #1123, opened 2023-12-14
test: added test for mergeAdditionalCallback

satyazzz123:
## test for mergeAdditionalCallback function 
i) this test is  for mergeAdditionCallBack to check for a nil output

fixes a part of #881

### DIFF
--- a/filesystem/merger_test.go
+++ b/filesystem/merger_test.go
@@ -93,3 +93,18 @@ func TestMergeProcessFileCallBack_SameContent(t *testing.T) {
 		}
 	})
 }
+func TestMergeAdditionCallBack(t *testing.T) {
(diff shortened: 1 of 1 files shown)


satyazzz123:
@seshapad @kmehant Please review this PR.

kmehant (commented):
Thanks for the contribution, @satyazzz123, At this point, we dont need a test case to cover this function as it simply returns nil. 

cc: @seshapad

HarikrishnanBalagopal:
> Thanks for the contribution, @satyazzz123, At this point, we dont need a test case to cover this function as it simply returns nil.
> 
> cc: @seshapad

Closing see above ^

PR #1179, opened 2024-04-22
fix: Checked if path is empty string or not

### LINKED ISSUE #1116: bug: doesn't handle the case where paths are empty properly
https://github.com/konveyor/move2kube/blob/e88ceceee7e8e76a587827d9555035566d4d0b38/common/utils.go#L1078

NeerajNagure:
Fixes #1116 
Checked is any one of path or parent is empty string and it returns false if its empty string

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -1106,6 +1106,9 @@ func ReverseInPlace[T interface{}](xs []T) {
 // Also returns true if the paths are the same.
 func IsParent(child, parent string) bool {
 	var err error
+	if parent == "" || child == "" { 
+		return false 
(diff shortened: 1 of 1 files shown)


kmehant:
@NeerajNagure Thank you for the PR. 

Did you come across any use case with Move2Kube tool that was passing empty strings to this function and triggered an error? If so can you point us to your use case? (input artifacts and exact command you used with Move2Kube tool)?

NeerajNagure:
@kmehant there was an issue open #1116 which I tried to fix. I have not came across any such use case.

PR #1186, opened 2024-07-05
feat: improve git clone mechanism

gabriel-farache:
Improving how git clone is performed:

- cleaning created local folder after 1st clone fails
- better log level to clearly inform the use of what's happening

Solves: https://github.com/konveyor/move2kube-api/issues/159


gabriel-farache:
On a side note: I find it strange and un-intuitive to force the creation of the source branch if it does not exists when cloning the repo, this should at least be better advertised (hence using warn instead of debug log level) and even, this behaviour should be explicitly enabled by a flag IMO (maybe it is but I could not find it)

seshapad:
@gabriel-farache could you please fix the build..seems like some of the libraries are affecting the license compliance.

kmehant (approved):
Thank you @gabriel-farache

PR #1188, opened 2024-07-22
test: cicd release failing with HTTP 404 error

HarikrishnanBalagopal:
https://github.com/konveyor/move2kube/actions/runs/10039196758/job/27744113100#step:4:259

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,38 +103,51 @@ jobs:
         with:
           github-token: ${{ secrets.MOVE2KUBE_PATOKEN }}
           script: |
+            core.info('DEBUG 1 start');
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -103,45 +103,60 @@ jobs:
         with:
           github-token: ${{ secrets.MOVE2KUBE_PATOKEN }}
           script: |
+            core.info('DEBUG 1');
```
seshapad, line 106: can we have some informative messages here? Applies to all the `core.info()` messages.

HarikrishnanBalagopal: done

seshapad (approved):

PR #1196, opened 2024-10-18
fix: podman doesn't support tags when using --cache-from

HarikrishnanBalagopal:


ashokponkumar:
can we make it conditional on the container runtime and have it for docker and not have it for podman?

PR #1197, opened 2024-10-18
fix: recover from panic in compose library parse function

HarikrishnanBalagopal:


transformer/compose/v1v2.go
```diff
@@ -124,6 +124,18 @@ func parseV2(path string, interpolate bool) (*project.Project, error) {
...
 
+// parseV2 parses version 2 compose files while capturing panics
+func parseV2(path string, interpolate bool) (result *project.Project, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			logrus.Errorf("recovered from panic in panicky_parseV2: %q", r)
+			err = fmt.Errorf("panicky_parseV2 failed: %q", r)
+		}
+	}()
+	result, err = panicky_parseV2(path, interpolate)
```
ashokponkumar, line 135: isn't it sufficient to wrap only `err = proj.Parse()` ?

HarikrishnanBalagopal: done

ashokponkumar (approved):

